PR #125, opened 2026-05-20
Fix cycle save MissingGreenlet serialization 500

jbk83:
## Summary
- refresh `Cycle` rows after `flush()` before serializing create/update responses
- prevents async SQLAlchemy from lazily loading expired `updated_at` outside greenlet context
- adds regression coverage that fails if update serializes the server-updated timestamp before refresh

## Validation
- `python -m py_compile brain/app/api/routers/cycles.py tests/test_cycles_service.py`
- manual async smoke of `update_cycle` with an expired-like `updated_at` sentinel

Note: pytest is not installed in this worker container (`No module named pytest`), so I could not run the full test file here.


redawear (approved):
Resolved the conflict with current main and re-ran the relevant backend checks. The MissingGreenlet refresh fix is still relevant.

PR #271, opened 2026-07-08
Widen minted App token to serve reads (contents + pull_requests read)

redawear:
## What & why

Completes the "`illo-bot` is Illo's GitHub identity" story on the **read** side. The App's installation token was minted **`issues:write` only**, so `read_github_source`'s `list_pull_requests` (and repo/code reads) got a **404 on the App token and fell back to a personal token** (Axel's / the static `GITHUB_TOKEN`). Net: `illo-bot` covered issue *authoring* but reads still rode on someone's personal PAT.

This widens the resolver's mint request to also grant **`contents:read` + `pull_requests:read`**, so the App serves those reads itself.

## Change

- `async_resolve_project_bound_env_tokens` (the App-minting site) now requests `{issues:write, contents:read, pull_requests:read}` instead of `issues:write` only.
- The App installation **already holds** these (granted at registration) → **no GitHub re-approval**.
- The added scopes are **read-only**, so the write blast-radius is unchanged — the token can still only *mutate* issues, never push code or open PRs.
- One token scope now serves both the App's reads and writes (shared cache entry per repo).

## What this does and doesn't cover

- ✅ **`read_github_source`** (list PRs, get repo, list issues) now resolves via the App — no more personal-token fallback for those.
- ⚠️ **`git clone` / file-contents in the exec sandbox** use a *separate* lane (the sync project-exec env, not this resolver), so they are **not yet** routed through the App. That's the deferred exec-lane wiring (E-exec-a). `contents:read` here is pre-positioned so the App token is ready the moment that lands.

## Testing

`240 passed` across `test_vault_project_tokens`, `test_github_app_mint`, `test_create_github_issue_tool`, `test_github_source_tool` (read path), `test_vault_hardening`, `test_runtime_secrets`, `test_agent_run_runtime`, `test_vault`, `test_vault_routes`, `test_tool_registry`, `test_builtin_skill_suite`. The resolver/contract assertions now expect the widened scope.

## Deploy note

Meant to ship **alongside #269** (already merged) in the next illo-dev deploy. After deploy, verify a `read_github_source list_pull_requests` on `uwear-backend` resolves `token_source: project_binding:GITHUB_TOKEN` with no `fallback_from_status_code`. Then the AXEL/static tokens are only needed for git-clone (exec lane) — safe to retire once that lane is wired or if clone isn't required.

### DIFF
--- a/brain/systems/vault/__init__.py
+++ b/brain/systems/vault/__init__.py
@@ -728,7 +728,16 @@ async def async_resolve_project_bound_env_tokens(
             env[env_name] = await async_mint_installation_token(
                 decrypted_blob,
                 repositories=[repo_name],
-                permissions={"issues": "write"},
+                # issues:write authors issues; contents+pull_requests read let the
+                # App also serve read_github_source (PR/repo reads) instead of 404
+                # falling back to a personal token. The App install already holds
+                # these (no re-approval) and they stay READ-only, so the write
+                # blast-radius is unchanged (the token can still only mutate issues).
+                permissions={
+                    "issues": "write",
+                    "contents": "read",
+                    "pull_requests": "read",
+                },
             )
         finally:
             decrypted_blob = None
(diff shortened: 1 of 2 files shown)


redawear:
Duplicate of #270, which was opened first by a parallel session and is more complete (it also widens DEFAULT_INSTALLATION_PERMISSIONS and updates test_github_app_mint). Closing in favor of #270 to avoid two competing PRs for the same change.

PR #348, opened 2026-07-17
Make cortex worker SIGTERM shutdown deterministic

redawear:
## Problem

The brain worker (`illospace-worker-1`) does not reliably exit on SIGTERM. Observed live on illo-dev 2026-07-16: after `docker kill -s TERM` from `update_worker_after_drain`, the worker lingered 2+ hours with **zero** active AgentRuns, and a follow-up `compose up -d --force-recreate` wedged 11+ minutes because compose honors the service's `stop_grace_period: 24h` while the process never exits.

Diagnosis (signal delivery itself is fine — python is PID 1 and a fresh idle worker exits in <1s):

1. `stop_runner(drain_timeout_seconds=None)` (prod: `ILLO_AGENT_RUNNER_DRAIN_TIMEOUT_SECONDS=infinity`) joined the `agent-runner-supervisor` thread with no bound. The supervisor's `asyncio.run()` cannot return while any slot task survives, and `brain/platform/async_io.py` deliberately awaits abandoned sync tool calls uninterruptibly — so one wedged no-timeout tool call from an already-settled run hangs shutdown forever while the DB shows 0 active runs.
2. Even when `main()` returns, interpreter shutdown joins the **non-daemon** `agent-tool` executor threads (`_TOOL_EXECUTOR`) forever if one is wedged.

## Fix

Drain what must drain, bound everything else, end with `os._exit`:

- **On TERM, stop claiming immediately**: signal handler calls new `request_runner_stop()`; `_run_queued_once_async` refuses to claim once stopping.
- **Drain waits on reality**: new in-flight run counter (also in `runner_health_snapshot`); `stop_runner` drains against it, honoring the drain timeout (`infinity` still never abandons active runs).
- **Teardown is always bounded**: supervisor join capped by `ILLO_AGENT_RUNNER_TEARDOWN_TIMEOUT_SECONDS` (default 10s, added to the compose worker env); a wedged supervisor is logged and skipped.
- **Exit is guaranteed**: `main()` ends every path with `os._exit(exit_code)` via a test-patchable hook after `logging.shutdown()` — wedged non-daemon threads can no longer block exit. Self-heal `SystemExit(1)` paths keep their exit code; startup crashes now log before exiting.
- **Cycle scheduler**: stop-aware sleep (wakes ≤0.5s) and `stop_cycle_scheduler` now joins the thread (bounded, warns on timeout) instead of only flipping a flag.
- **Deploy script**: `wait_for_worker_exit` prints a one-time hint when the worker has 0 active AgentRuns but won't exit (instead of silently waiting up to 24h), with the force-replacement recipe. Wait semantics unchanged.

## Tests

38 pass across `tests/test_cortex_worker.py`, `tests/test_cycle_scheduler.py`, `tests/test_agent_run_runner.py` — including: bounded `stop_runner` return with a wedged supervisor thread, drain waiting on in-flight runs, claim refusal while stopping, in-flight counter registered before the claim UoW exits (race coverage), prompt scheduler-thread join, terminate-hook invocation with correct exit codes, and signal-handler behavior. `bash -n` passes on the shell change.

## Verification plan

Pre-merge on illo-dev (branch build): `docker kill -s TERM` on an idle worker must exit within seconds. Result will be posted below.


redawear:
**Pre-merge verification on illo-dev (2026-07-17 00:42Z):** branch image built on the server, both 0026 migrations applied cleanly after the re-chain (0025 → packet_brief_deliveries → chantier), idle worker replaced onto the branch image, then `docker kill -s TERM`:

- **Process exited in 5.8s, exit code 0** (previously: 2+ hours / never, observed same day).
- Log sequence as designed: `received SIGTERM, draining` → bounded teardown (cycle-scheduler join hit its 5s cap and proceeded with a warning — the graceful path working) → `agent-run worker stopped` → hard exit.
- Worker restarted healthy afterwards; runner + cycle scheduler up.

Noted during verification (pre-existing, filed separately): a transient boot-time `RuntimeError: Lock bound to a different event loop` from SQLAlchemy's engine-event exec_once mutex — the worker's multiple event loops share one engine's event state. Unrelated to this diff.

PR #442, opened 2026-07-23
fix(scheduler): await the meta-evolution coroutine in the nightly step (#441)

### LINKED ISSUE #441: nightly_sleep: un-awaited coroutine crashes meta-evolution step; failure alert never fires; memory_health_log still empty
Found during a read-only illo-dev inspection (2026-07-23) while verifying the "dreams" pillar for #440.

**What works now** (post #341–#347 / #385 lineage): the scheduler catalog syncs (2 jobs registered 2026-07-17 12:13), and reconstructive consolidation persists and completes nightly — the varchar(20) truncation is clearly fixed:

```
 id |  run_date  |            phase             |  status   |         started_at
  4 | 2026-07-22 | reconstructive_consolidation | completed | 2026-07-23 07:00:01
  3 | 2026-07-21 | reconstructive_consolidation | completed | 2026-07-22 07:00:11
  2 | 2026-07-20 | reconstructive_consolidation | completed | 2026-07-21 07:00:05
  1 | 2026-07-19 | reconstructive_consolidation | completed | 2026-07-20 07:00:07
```

**What's broken:**

1. `scheduler_jobs` row for `nightly_sleep` has `consecutive_failure_count = 2` with `last_failure_error`:
   ```
   File "<string>", line 1, in <module>
     from brain.systems.feedback.meta_evolution import run_meta_evolution; stats = run_meta_evolution(); print(f'Insights: {stats["insights_total"]}, ...')
   TypeError: 'coroutine' object is not subscriptable
   RuntimeWarning: coroutine 'run_meta_evolution' was never awaited
   ```
   The night-budget step invokes `run_meta_evolution()` synchronously from an inline `python -c`-style string, but it is an async coroutine. Fix: await it (`asyncio.run(...)`) or make the step handler async-aware.

2. `failure_alerted_at` is **NULL despite 2 consecutive failures** — the repeated-failure alert (the #385 ask) either has a threshold above 2 or doesn't fire. A failing nightly_sleep is currently invisible to the team, which is exactly the silent-death class this alert exists to prevent (66h cycle-wedge precedent).

3. `memory_health_log` still has **0 rows ever** — the health phase is presumably never reached after the crash. Verify it produces rows once (1) is fixed.

**Acceptance:** nightly_sleep completes end-to-end (meta-evolution stats logged), a forced failure demonstrably fires the alert path, and `memory_health_log` gets its first rows.

redawear:
Closes #441

## What was broken

The `meta_evolution` nightly step built an inline python one-liner that called `run_meta_evolution()` **synchronously**, but that function is `async def` (`brain/systems/feedback/meta_evolution.py:305`). So the step raised `TypeError: 'coroutine' object is not subscriptable` on every run — which is verbatim what production's `scheduler_jobs.last_failure_error` holds for `nightly_sleep`.

The existing tests in `test_meta_evolution.py` all `await` it correctly, which is why nothing caught the scheduler call site.

## The fix

Extracted `nightly_meta_evolution_command()` wrapping the call in `asyncio.run(...)`. This **follows the existing `nightly_heuristic_review_command()` pattern** in the same module rather than introducing a second shape for "nightly step that runs a coroutine".

## Two of the issue's three claims were misdiagnosed — read before merging

The issue was filed from a production inspection and guessed at causes. Verified against the code:

**Item 2 — the NULL alert is CORRECT, not a bug.** The issue saw `consecutive_failure_count = 2` with `failure_alerted_at` NULL and guessed "threshold above 2 or doesn't fire". `SCHEDULER_FAILURE_ALERT_THRESHOLD_DEFAULT = 3` (`brain/app/scheduler/runtime.py:39`) and the guard fires exactly once at the threshold. **The threshold is deliberately left at 3 here.** This PR only adds a test proving the alert edge sequence is `[False, False, True, False]` and resets on success.

⚠️ There *is* a real gap underneath, deliberately **not** fixed here: when the alert does fire, its only consumer (`brain/app/scheduler/executor.py:403`) emits a `logger.error()` and nothing else — no Slack, no thread post, no Illo record. A nightly job can fail forever with no one told. Where that alert should land is a product decision, raised separately.

**Item 3 — `memory_health_log` is empty for a different reason.** The issue assumed the health phase "is never reached after the crash". In fact `MemoryHealthRepository.record()` has **no production caller at all** (only the model, the repository, and two tests). **Fixing this crash will not produce rows.** Nothing was invented to fill the table; whether a nightly memory-health phase should exist is a design question, not a bug fix.

## Verification

```bash
cd <worktree> && /Users/redamjahed/Documents/UwearDev/illospace-project/venv/bin/python \
  -m pytest tests/test_scheduler.py tests/test_meta_evolution.py -q
```
**59 passed.**

**Negative control** (the test has teeth): reverting `programs.py` and re-running the new test reproduces the exact production error —
`TypeError: 'coroutine' object is not subscriptable`. The test executes the production-composed command string rather than grepping it for `asyncio.run`.

## Acceptance checks

1. `nightly_sleep`'s `meta_evolution` step completes and prints its stats line instead of raising.
2. The regression test fails without the fix and passes with it (verified above).
3. The repeated-failure guard opens exactly one alert at the 3rd identical failure and resets after success — threshold unchanged at 3.
4. No unrequested behaviour change to the scheduler, the guard, or memory health.

### DIFF
--- a/brain/app/scheduler/programs.py
+++ b/brain/app/scheduler/programs.py
@@ -29,6 +29,18 @@ def nightly_heuristic_review_command() -> list[str]:
     )
 
 
+def nightly_meta_evolution_command() -> list[str]:
+    """Run async meta-evolution from a synchronous scheduler process."""
+    return _python_one_liner(
+        "import asyncio; "
+        "from brain.systems.feedback.meta_evolution import run_meta_evolution; "
+        "stats = asyncio.run(run_meta_evolution()); "
+        "print(f'Insights: {stats[\"insights_total\"]}, "
+        "Regressions: {stats[\"regressions\"]}, "
+        "Adjustments: {len(stats[\"adjustments\"])}')"
+    )
+
+
 @dataclass(frozen=True)
 class StepSpec:
     step_key: str
@@ -134,13 +146,7 @@ NIGHTLY_STEP_REGISTRY: tuple[NightlyStepDefinition, ...] = (
     ),
     NightlyStepDefinition(
         step_key="meta_evolution",
-        command_factory=lambda _target_date: (
-            _python_one_liner(
-                "from brain.systems.feedback.meta_evolution import run_meta_evolution; "
-                "stats = run_meta_evolution(); "
-                "print(f'Insights: {stats[\"insights_total\"]}, Regressions: {stats[\"regressions\"]}, Adjustments: {len(stats[\"adjustments\"])}')"
-            ),
-        ),
+        command_factory=lambda _target_date: (nightly_meta_evolution_command(),),
         description="Meta-evolution",
         budget_hint={
             "work_type": "context_policy_eval",
--- a/tests/test_scheduler.py
+++ b/tests/test_scheduler.py
@@ -581,6 +581,59 @@ async def test_nightly_heuristic_review_wrappers_await_and_report_counts(monkeyp
     ) == [command]
 
 
+async def test_nightly_meta_evolution_wrapper_awaits_and_reports_stats(monkeypatch, capsys):
+    from brain.systems.feedback import meta_evolution
+
+    awaited = False
+
+    async def fake_run_meta_evolution():
+        nonlocal awaited
+        awaited = True
+        return {
+            "insights_total": 3,
+            "regressions": 1,
+            "adjustments": {"heuristic_prune_threshold": {"new": 0.15}},
+        }
+
+    monkeypatch.setattr(
+        meta_evolution,
+        "run_meta_evolution",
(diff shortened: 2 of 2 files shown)


redawear:
Superseded by #445, which merged the identical fix — `main` now has `nightly_meta_evolution_command()` wrapping `run_meta_evolution()` in `asyncio.run(...)`. This PR is now CONFLICTING against `main` and has nothing left to add. Recommend closing it unmerged rather than resolving the conflict.

(Flagged by the R3 routine, which had this open as a draft when #445 landed.)

PR #470, opened 2026-07-24
fix(routing): move the fleet default back to GPT-5.5 on latency grounds

redawear:
**Ready to merge on your word — this reverses your "only 5.6 SOL" call, so I'm not merging it unilaterally.** Data first.

## What happened
After the routing deploy moved the org default to gpt-5.6-sol, Illo stopped completing work. Throughput per hour, from `agent_runs`:

| hour | created | completed |
|---|---|---|
| 16:00 | 4 | 4 |
| 17:00 | 7 | 7 |
| 18:00 | 7 | 7 |
| 19:00 (deploy) | 13 | 4 |
| 20:00 | 5 | **0** |

13 active runs, 8 queued and growing. Runs were alive and progressing — no errors, zero failed API calls — just far too slow, so the worker's slots stayed occupied for 40–80 minutes per run and everything else queued behind.

## Cause: 5.6-sol is 2–4x slower per call
From `agent_api_calls` (last 3h):

| model | effort | avg latency |
|---|---|---|
| gpt-5.5 | high/medium | **~18–20s** |
| gpt-5.6-sol | low | 43.4s |
| gpt-5.6-sol | medium | 72.4s |
| gpt-5.6-sol | high | 78.5s |

Run 2557 straddled the deploy and isolates the model as the only variable: **5 calls on 5.5 at 17.9s, then 43 calls on 5.6-sol at 53.9s for the same task.** At 20–40 turns per run, that is the whole outage.

**Effort routing cannot fix this**: 5.6-sol at `low` (43s) is still slower than 5.5 at `high` (20s). It's a model property, not a tier property — 5.6-sol is a preview model (the composer literally labels it "Preview model for eligible personal connections") and is very likely capacity-limited.

## What this changes
Only the *default model*. Migration 0040 moves org config back to `openai/gpt-5.5` and the code default follows. **The routing layer is untouched and still correct** — cycle overrides honored, canonical effort tiers, provider-generic rendering, one-model principle. 5.6-sol stays fully selectable per cycle, per run, and in the composer, so you can put it on judgment-heavy lanes where 80s/call is fine.

Recommendation: merge this to restore throughput, then decide whether 5.6-sol goes on cycle 2 (Coordinator) specifically, where quality matters more than latency. Revisit the fleet default when 5.6-sol's latency lands near 5.5.

Tests updated for the default flip: 176 passed across model_policy/providers/predict_rlm/effort_rendering/cycles_service.

### DIFF
--- a/brain/platform/db/alembic/versions/0040_default_model_latency_rollback.py
+++ b/brain/platform/db/alembic/versions/0040_default_model_latency_rollback.py
@@ -1,0 +1,48 @@
+"""Move the fleet default back to GPT-5.5 on latency grounds.
+
+GPT-5.6 Sol is a preview model and is 2-4x slower per call than GPT-5.5 at
+every effort tier (measured on illo-dev 2026-07-24: 5.5 ~18-20s; 5.6-sol
+43s at low, 72s at medium, 79s at high). Runs average 20-40 turns, so the
+fleet default at 5.6-sol pushed run duration past 40 minutes, saturated the
+worker, and dropped completions to zero while the queue grew.
+
+Effort routing cannot recover this: 5.6-sol at `low` is still slower than 5.5
+at `high`. The routing layer itself is unchanged and correct — only the model
+the fleet defaults to moves back until 5.6-sol is fast enough to carry it.
+
+Revision ID: 0040_default_model_latency_rollback
+Revises: 0039_clear_revision_model_pins
+Create Date: 2026-07-24
+"""
+
+from __future__ import annotations
+
+from alembic import op
+
+
+revision = "0040_default_model_latency_rollback"
+down_revision = "0039_clear_revision_model_pins"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.execute(
+        """
+        UPDATE orgs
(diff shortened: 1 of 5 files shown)


redawear:
Closing — the premise is disproven. I compared gpt-5.5 on an idle system against gpt-5.6-sol on a saturated one, which is not a valid comparison.

Controlled A/B on illo-dev, same request, same effort, back to back:

| context | gpt-5.5 | gpt-5.6-sol |
|---|---|---|
| minimal | 1.3s | 1.3s |
| ~30k tokens, medium | 2.6s | **2.2s** |
| ~30k tokens, high | 2.7s | **1.7s** |

5.6-sol is equal or **faster**, exactly as Reda said. The production 55-70s calls are heavy chantier/PR-review turns (`turn 32, msgs=7, tools=83`, dozens of GitHub blob fetches) — genuine reasoning time on genuinely heavy work, not a model regression. The org default stays gpt-5.6-sol.

The investigation did surface a real defect, fixed separately: every OpenAI Codex request sends `prompt_cache_retention`, which that backend rejects with a 400, so each call pays a wasted round trip and never gets the intended 24h cache retention.

PR #531, opened 2026-07-27
fix(alembic): re-chain 0044 onto 0043 — main has two heads

redawear:
## What this is

`main@3b116ce` has **two alembic heads**, so `alembic upgrade head` is ambiguous and a fresh-DB migrate fails. `#521` (cycle liveness) and `#523` (deferral obligations) each chained a migration off `0042`:

```
0043_cycle_liveness_controls       (head) -> 0042
0044_open_ask_deferral_obligations (head) -> 0042
```

This re-chains `0044` onto `0043` — **both** the `Revises:` docstring line and `down_revision`. The two migrations are independent (`0043` touches cycle fields, `0044` touches `open_asks`/`obligation_notices`), so the order is arbitrary and re-chaining is safe. One-line-each change, no schema or behavior difference.

## The guard already existed — that is the interesting part

`tests/test_alembic_config.py::test_alembic_revision_chain_has_single_head` has been asserting exactly one head all along, and it **fails correctly** on current `main`. I briefly added a second guard before finding it; that duplicate is removed in the second commit here.

So the gap is not missing coverage — it is that **each PR passed the guard alone, and nothing re-ran it against the updated `main` at merge time.** Both branches were cut before the other's migration existed. Any two PRs that add a migration will do this again; only a merge queue (or a required re-run against the merge target) closes it. Flagging rather than fixing — that is a repo-settings call, not mine.

## Verification (executed)

```bash
cd /Users/redamjahed/Documents/UwearDev/illospace-project
./venv/bin/python -m alembic heads
./venv/bin/python -m pytest tests/test_alembic_config.py -q
```

- `alembic heads` on this branch: one head, `0044_open_ask_deferral_obligations`. On `main`: two.
- `tests/test_alembic_config.py`: **14 passed** here; the single-head case **fails on `main`**.
- Full suite on the merge of this branch's fix + PR #526: **4264 passed, 76 skipped, 0 failed**. Before the fix that same tree was `1 failed` — the head assertion, and nothing else.

## Acceptance checks

1. `./venv/bin/python -m alembic heads` prints exactly one head.
2. `test_alembic_revision_chain_has_single_head` passes here and fails against `main@3b116ce`.
3. `0044`'s docstring `Revises:` and its `down_revision` both read `0043_cycle_liveness_controls` — fixing only one leaves `test_alembic_revision_headers_match_identifiers` red.
4. No schema change, no data migration, no behavior change.


redawear:
Superseded by #532, which you merged at 18:50:46Z with the identical two-line re-chain. I opened this ~2 minutes later without re-checking `main` — my own freshness-check failure, not a disagreement about the fix.

Closing unmerged. `main@fefd4f9` is single-head and verified: `alembic heads` returns `0044_open_ask_deferral_obligations` alone, and the full suite runs 4319 passed / 0 failed on a branch cut above it.

The one thing here that #532 does not cover, and that I'd still like an answer on: **the single-head guard already existed** (`tests/test_alembic_config.py::test_alembic_revision_chain_has_single_head`) and passed on both #521 and #523, because each branch was cut before the other's migration existed. No test can close that — only a merge queue, or requiring the check to re-run against the merge target. Raised as a DECIDE in today's R3 report.

Issue #497, opened 2026-07-25
[scheduler][guard] Repeated-failure alert misses intermittent failures — 7 failures in 15h, zero alerts (streak resets on interleaved successes)


redawear:
## The gap, observed live overnight (2026-07-24→25)

`uwear_aws_health_scan` failed **7 of 15** scheduled runs — a 47% miss rate — and the repeated-failure guard (#441/#445) never alerted, because successes were interleaved: `async_reset_scheduler_job_failure_guard` zeroes `consecutive_failure_count` on every success, so the streak never reached the threshold. `failure_alerted_at` is NULL; `consecutive_failure_count` is 0 right now despite 7 failures in 15 hours.

The guard as built catches **sustained death**. It is blind to **intermittent death**, which for an hourly monitor is nearly as bad: every missed hour is a blind window (see the companion reliability ticket), and nobody is told.

## Change

Add a second, independent trigger to the failure guard: alert when a job accumulates **N failures within a rolling window** (default: 3 failures in 24h), regardless of interleaved successes.

- Count from `scheduler_runs` (`status = 'settled_failure'`, `started_at > now - window`) at failure-record time — no new tables, no new counters.
- One-shot semantics like the existing edge: after firing, don't re-fire until the rolling count drops below threshold and crosses again (a `rate_alerted_at` timestamp on `scheduler_jobs`, mirroring `failure_alerted_at`, is enough; reset alongside the existing guard reset only when the rolling window is clean).
- Reuse the existing Slack delivery path (`async_deliver_scheduler_failure_alert`) with a message variant: "N failures in the last 24h (intermittent)" + job key + last error head.
- Config: `SCHEDULER_FAILURE_RATE_THRESHOLD` / `SCHEDULER_FAILURE_RATE_WINDOW_HOURS` env with safe defaults (3 / 24). No feature gate — always on.

## Acceptance

- Simulated alternating success/failure reaching 3 failures in the window ⇒ exactly one Slack alert; further alternation without recovery ⇒ no re-fire spam.
- Sustained-streak behavior (existing threshold edge) unchanged; both triggers can coexist on one job.
- Existing tests green; new tests for the rolling counter and the one-shot semantics.

Live proof note: the streak-edge alert has still never fired in production (#452 was fixed before the third identical nightly failure). When this lands, its first live firing will double as the delivery-path production test — 7-failure nights like last night would have triggered it.

PR #534, opened 2026-07-27
refactor(scheduler): one failure guard with a real trigger registry

### LINKED ISSUE #499: [scheduler][refactor] Unify the failure guard's two triggers behind one evaluation, instead of a second guard bolted beside the first
Follow-up from the thermo-nuclear maintainability review of #497 (PR opened by R3). The #497 change is correct and shippable — the review confirmed all three of its correctness questions — but it was constrained by that ticket's explicit "no new tables, no new counters" rule, so it adds a *second* guard beside the existing one rather than modeling one guard with multiple triggers. This ticket is the structural cleanup that constraint deferred.

## The finding

The distinction between the two triggers is semantically justified — the consecutive trigger resets on signature change or success, while the rolling trigger stays armed across successes until its window cools. The problem is that this distinction is now re-encoded at every layer:

- Two unrelated `*_alert_emitted` booleans in an untyped dict returned by `async_record_scheduler_job_failure`.
- `async_deliver_scheduler_failure_alert` takes a bundle of mutually-exclusive optional arguments and infers which alert it is from whichever value happens to be non-null.
- `_async_apply_failure_guard` duplicates the whole log-then-deliver branch.
- The canonical scheduler serializer and the health snapshot are still consecutive-only, so one representation of "the failure guard" knows about the rolling latch and another does not.

## Proposed shape

- A typed `FailureGuardEvaluation` holding guard state plus a collection of `FailureGuardEdge` values discriminated by kind (`consecutive` / `rolling_window`), each carrying its count, threshold, optional window, latch timestamp, and whether it crossed.
- Evaluate all configured triggers under the existing job lock, then send the resulting edges through **one** delivery/coalescing path. Simultaneous crossings should produce a deliberate combined notification rather than today's `elif` suppression.
- Persist latches keyed by trigger kind — a normalized `(job_id, trigger_kind, alerted_at)` table — instead of one nullable column per trigger. **This is the part #497 was explicitly forbidden from doing**, and it is the reason this is a separate ticket.
- One public serializer shared by run summaries, catalog output, and health alerts, so a third trigger cannot be half-visible again.
- Load the settings into one failure-guard policy object rather than a getter pair per trigger.

## Also deferred from that review (non-blocking)

- A public-path integration test that drives hourly runs through `async_run_scheduler_run` with `max_attempts=2` and asserts the rolling count after both attempts. The current rate tests build settled rows by hand and call the private helper, so they cannot catch production sequencing or retry double-counting. (Both were verified by hand in review: the current run *is* flushed before evaluation, and a retried hour contributes exactly one row.)
- `tests/test_scheduler.py` grew from 1,672 to 1,867 lines; the failure-guard suite wants its own module.

## Acceptance

- One evaluation path produces every alert edge; adding a third trigger touches the trigger list and nothing else.
- Simultaneous crossings deliver one message by explicit design, and no return value claims an alert was emitted when none was.
- Serializer, catalog output, and health snapshot all report every trigger's state.
- Existing consecutive-streak and rolling-window behavior unchanged; #497's tests still pass.

<sub>Filed by Routine R3 from the Codex thermo-nuclear review of #497, run 2026-07-25 12:05 ET.</sub>

redawear:
Closes #499

## What was wrong

The failure guard has two triggers — a consecutive-failure streak and a rolling window — and the distinction between them was re-encoded at every layer: two unrelated `*_alert_emitted` booleans in an untyped dict, a delivery helper that sniffed mutually-exclusive optional arguments to work out which alert it held, a duplicated log-then-deliver branch, and a serializer plus health snapshot that knew only about the consecutive trigger. One representation of "the failure guard" knew about the rolling latch and another did not.

## What this does

One guard, a registry of triggers, one evaluation and one delivery path. Each trigger owns its configuration, evaluation, public details, alert text and reset predicate; the engine knows only how to load latches, detect crossings and collect edges. `runtime.py` drops from 993 to 592 lines with the subsystem extracted to `failure_guard.py`.

Latches move from one nullable column per trigger to a normalized `(job_id, trigger_kind, alerted_at)` table — **the thing #497 was explicitly forbidden from doing, and the reason this was a separate ticket** (migration `0045`, chained off `0044`, single head verified).

Simultaneous crossings now coalesce into one message by design, instead of an `elif` silently suppressing one.

## Two things to know before merging

**1. External payload shape changed.** `failure_guard` goes from flat consecutive/rate-specific fields to a `triggers` collection, and health alerts use `type: "scheduler_job_failure_guard"` with the same collection. That is required by the ticket's acceptance ("serializer, catalog output and health snapshot all report every trigger's state") — but it is a real payload change, not a refactor, so anything parsing the old flat fields needs to know.

**2. One genuine behavior change beyond reporting.** A **rolling-only** latch now appears in health *and degrades health status*; previously only the consecutive trigger could. That means a job can newly reach a degraded state it never reached before. I judged it in scope — a trigger that can fire but cannot affect health is exactly the half-visibility this ticket exists to remove — but it is the line between "report every trigger" and "act on every trigger", so it is your call.

## Verification (executed on this branch)

```bash
cd /Users/redamjahed/Documents/UwearDev/.codex-worktrees/illospace-499-fix1
/Users/redamjahed/Documents/UwearDev/illospace-project/venv/bin/python -m pytest -q
/Users/redamjahed/Documents/UwearDev/illospace-project/venv/bin/python -m alembic heads
```

**Full suite: 4321 passed, 76 skipped, 0 failed.** `alembic heads`: one head. (Run outside the worker sandbox, where six GPU/OAuth tests fail on `PermissionError` binding `127.0.0.1`.)

## Acceptance checks — each a named test I ran individually

1. **"Adding a third trigger touches the trigger list and nothing else"** is now demonstrably true, not asserted: `test_third_trigger_flows_through_engine_serializer_health_and_delivery` registers a runtime-duration trigger with `elapsed_minutes`/`minimum_runtime_minutes` — **no count/window contract at all** — and drives it through evaluation, persistence, serialization, health and Slack delivery without touching engine code.
2. The migration round-trips **all four** latch states exactly: `test_failure_guard_migration_restores_exact_prior_schema[consecutive-only | rolling-only | both | neither]`.
3. A downgrade that would silently lose a trigger kind it cannot represent **raises instead**: `test_failure_guard_downgrade_rejects_unrepresentable_trigger`.
4. `tests/test_alembic_config.py` — 14 passed, including the single-head guard and the reviewed-destructive-migration allowlist that `0045`'s new `drop_table` is registered in.

## Review history

Built, then put through a cross-family code-quality review, which returned three blocking findings. The sharpest was a real defect, not a style call: `downgrade()` recreated the two legacy columns but **never dropped the new table**, so a downgraded database matched neither revision. The review also disproved the ticket's own extensibility claim — `window_hours` was hard-coded through the supposedly generic contract. Both are fixed in the second commit, along with splitting the ~590-line guard suite out of `test_scheduler.py` (2203 → 1520 lines).

## Follow-up filed separately

The review's one non-blocking finding — catalog and health load latches and run every trigger counter per job sequentially, so each new DB-backed trigger adds a query per job — is deliberately **not** in this PR. It is a performance change on top of an already-large refactor.


redawear:
Superseded by #537 — closing this to keep one PR per issue.

Both PRs say `Closes #499`, and this branch (`illo-qa/499-fix1`) is a **strict ancestor** of #537's `illo-qa/499-fix2`:

```
git merge-base --is-ancestor origin/illo-qa/499-fix1 origin/illo-qa/499-fix2   # true
git log --oneline illo-qa/499-fix1..illo-qa/499-fix2                           # 2dbac5d test(scheduler): prove failure guard production wiring
```

So #537 contains everything here plus the fix pass that came out of the second code-quality review. Merging this one instead would have shipped #499 without it.

My mistake for opening a second PR for the fix pass rather than pushing it onto this branch. #537 is the one to review.

PR #543, opened 2026-07-27
test(cycles): prove the wake lock on Postgres — and fix the three guards it was not protecting

redawear:
Closes the minor finding deferred from the adversarial review of #536: the wake
tests all ran on one shared SQLite session, where `SELECT ... FOR UPDATE` is a
no-op, so nothing proved that `async_wake_cycle_now`'s lock actually ordered
competing callers.

It turned out not to be a documentation exercise. **The lock was real but the
guards behind it were not** — and the same defect appears three times in
`brain/systems/cycles/service.py`, once in a path that can execute a cycle run
twice.

## The defect

`SELECT ... FOR UPDATE` re-reads a row the unlocked lookup already put in the
session's identity map, and SQLAlchemy returns that **cached pre-lock instance**
rather than repopulating it from the row it just locked. Postgres blocks the
second caller correctly; the second caller then decides everything from state
its competitor has already replaced. Verified directly — raw SQL in the same
transaction saw the new value while the ORM object still held the old one.

| where | what it meant |
| --- | --- |
| `async_wake_cycle_now` | two concurrent wakes both returned `woken` and both stamped `next_run_at`; the `enabled` / `deleted_at` re-check the lock exists to protect never re-checked anything |
| `async_claim_cycle_run` | two executors both claimed the **same** queued `CycleRun` and both returned `running` — the same cycle run can execute twice |

Two smaller holes in the same guard:

- `now` was sampled at function entry, so a caller that spent time queued on the
  lock judged the slot against a clock reading from before it waited.
- The locking re-read matched on `id` alone. A cycle renamed while a wake waited
  was woken anyway, under a name nobody asked for.

Each was reproduced against Postgres before being changed. No duplicate
`CycleRun` was ever produced by the wake path itself — #536's core safety claim
holds — but the dispositions callers log and act on were wrong.

## The fix

`populate_existing=True` on the locking selects, the selector predicates
re-applied to the locked row (Postgres re-evaluates them against the version it
waited for, so a row that stopped matching drops out), and the clock sampled
under the lock. Rebased onto #541, which landed the same `_aware_utc` normalization
independently; its `pending_at` naming is kept. Two fast-suite assertions moved
for the same reason #541 describes — they were comparing an in-memory object
the old code never refreshed, and now the value genuinely comes off the row.

Production semantics are otherwise unchanged.

## The tests

`tests/test_cycles_wake_concurrency.py`, 9 tests in the `requires_db` lane, each
caller on its own committed transaction. The four cases asked for, plus the
three regressions above.

Determinism comes from staging the contention rather than hoping for it: an
outside connection pins the row, or the scheduler is paused mid-claim at a real
seam, and the test proceeds only once `pg_blocking_pids` confirms its own
backends are actually blocked. Nothing waits on a guessed sleep.

Each test owns a **private schema**. The scheduler selects every due cycle in the
database, so committed rows from anywhere else silently steer these tests — not
hypothetical, an early teardown failure here left orphan due cycles behind and
the suite then failed two different ways depending on what got picked up.
Engines bind to `TEST_DB_URL`, not `config.DB_URL`, which prefers an inherited
`DATABASE_URL` and could aim committed writes at a developer's own database.

## Verification

- Full `requires_db` suite against a dockerized pgvector pg16: **71 passed**
  (62 before this branch), no orphan rows or schemas left behind.
- Full fast suite: **4327 passed**, 6 skipped.
- The concurrency module 20x consecutively under sustained CPU saturation, after
  a load-dependent flake was found and fixed (an every-minute cron let the
  recomputed backstop fall due before the blocked wake got the row, at which
  point `already_pending` was the honest answer and the test was asserting a
  state it had not pinned).

Commits are ordered failing-test-then-fix, so each race is demonstrated before
it is closed.

## A limit on what this lane can prove

Worth stating plainly, because this PR's premise is that Postgres tests catch
what SQLite cannot. They do — but not everything.

`0001_public_schema_baseline.py` has no frozen DDL: its `upgrade()` calls
`Base.metadata.create_all(bind)`, materializing whatever the models say *today*.
A freshly migrated database therefore gets today's types, while production got
the types the models had when it was built. Where a model changed type without
an explicit `ALTER` migration, the two diverge permanently and silently.

`cycles.next_run_at` is exactly that. The model says `DateTime(timezone=True)`;
a freshly migrated pg16 gives `timestamp with time zone` and asyncpg returns
aware datetimes (verified — it is why the tests here compare aware values and
pass). Production returns naive, which is the live TypeError #541 fixed. And
`grep -rn "next_run_at" brain/platform/db/alembic/versions/*.py` returns
nothing: no migration ever created or altered that column.

So no test in this lane could have caught #541's bug, because the test database
is structurally the *correct* schema and production is not. That
`0002_idea_timestamps_timestamptz` and `0003_agent_run_timestamps_timestamptz`
exist suggests this was swept for two tables and not the rest. Filed as a
follow-up; nothing here depends on it.

## Follow-up, not in this PR

`brain/systems/cycles/exception_ping.py` repeats the same lock-then-read-stale
pattern (~lines 428/434, 564/570, 615/621). Lower impact — a later caller can
overwrite `context_snapshot` from stale state and lose an earlier audit decision
— but it is the same mechanism and wants its own change with its own tests.

Findings 1, 2, 3, 4 and 6 originate from a cross-family Codex review of this
branch; each was independently reproduced before being acted on.

### DIFF
--- a/brain/systems/cycles/service.py
+++ b/brain/systems/cycles/service.py
@@ -406,8 +406,17 @@ async def async_claim_cycle_run(session, run_id: int) -> tuple[CycleRun, Cycle] 
     if not candidate or candidate.status != "queued":
         return None
 
+    # Both locking reads need populate_existing, for the same reason the wake
+    # primitive does: the unlocked read above cached this run in the identity
+    # map, and the ORM would hand that copy back rather than the row it just
+    # locked. The Cycle lock guarantees a second executor arrives only after the
+    # first committed — so without the refresh it reads "queued", claims a run
+    # that is already running, and executes it twice.
     result = await session.scalars(
-        select(Cycle).where(Cycle.id == candidate.cycle_id).with_for_update()
+        select(Cycle)
+        .where(Cycle.id == candidate.cycle_id)
+        .with_for_update()
+        .execution_options(populate_existing=True)
     )
     cycle = result.first()
     result = await session.scalars(
@@ -417,6 +426,7 @@ async def async_claim_cycle_run(session, run_id: int) -> tuple[CycleRun, Cycle] 
             CycleRun.cycle_id == candidate.cycle_id,
         )
         .with_for_update()
+        .execution_options(populate_existing=True)
     )
     run = result.first()
     if not run or run.status != "queued":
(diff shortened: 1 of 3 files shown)


redawear:
Context: this closes the minor finding deferred in the adversarial review of #536 ("the wake tests run on a single shared SQLite session where FOR UPDATE is a no-op, so real lock-ordering behavior is unproven").

The finding was correct, and following it up found more than unproven behavior — the lock was real but three guards behind it were reading pre-lock state, including one (`async_claim_cycle_run`) that can execute the same CycleRun twice. Details in the PR description.

PR #559, opened 2026-07-28
refactor(scheduler): let failure-guard triggers own mutable persisted state

### LINKED ISSUE #538: [scheduler][refactor] Failure-guard triggers cannot own mutable persisted state — the engine still runs the consecutive-failure counter
Follow-up from the second code-quality review of #499 (PR https://github.com/Illospace/illospace/pull/537). Filed rather than fixed inside #499 because closing it means a second schema change on top of the `0045` table that PR already adds, and #499 does not ask for it.

## The finding

`brain/app/scheduler/failure_guard.py` — after #499, a trigger owns its configuration, evaluation, public details, alert text and reset predicate, and #537 proves end to end that a **read-only** third trigger registers through the production registry with **zero engine changes** (`test_third_trigger_flows_through_production_apply_health_delivery_and_reset`).

But the engine still owns the `consecutive_failure_count` state machine and its success reset. So the extensibility guarantee holds only for triggers that derive their verdict from data the engine already gathers. **A trigger that needs its own mutable persisted state still requires engine changes** — which means #499's headline acceptance line ("a third trigger touches the trigger list and nothing else") is true for the read-only case and false for the stateful one.

## Proposed shape

- Move failure accumulation and reset behind trigger lifecycle methods, so a trigger declares the state it owns rather than the engine hard-coding one counter.
- Generic per-trigger state persistence — the `scheduler_failure_guard_latches` table introduced by `0045` is the natural place, and extending it is a schema change, hence the separate ticket.
- Prove it the same way #537 did: a third trigger that owns custom **mutable** state, exercised through the production registry across apply → snapshot → delivery → reset.

## Acceptance

- A stateful third trigger is added by touching the trigger registry and nothing in the engine.
- The existing consecutive and rolling-window triggers keep byte-identical behavior.

<sub>Filed by Routine R3 from the second Codex code-quality review of #499, 2026-07-27.</sub>

redawear:
Closes #538.

## What it does

After #499, a failure-guard trigger owned its configuration, evaluation, public details, alert text and reset predicate — and #537 proved a **read-only** third trigger could be registered with zero engine changes. But the engine still hard-coded the `consecutive_failure_count` state machine and its success reset. So the extensibility guarantee was true for the read-only case and false for the stateful one: any trigger needing its own mutable persisted state still meant editing the engine.

Failure accumulation and reset now sit behind trigger lifecycle methods, with generic per-trigger state persistence. A stateful third trigger is added by touching the trigger registry and nothing else — proven end to end through the **production** registry across apply → snapshot → delivery → reset.

## ⚠️ This migration drops two populated columns

`0047` creates dedicated trigger-state tables, backfills `consecutive_failure_count` into `{"count": ...}`, and then **drops the legacy count columns** from both the scheduler and cycle guards. That is deliberate — keeping them was a dual write with two sources of truth — but it is a one-way data change on deploy, so it deserves your eye rather than mine.

The downgrade validates all state *before* mutating, restores representable counts, and **explicitly rejects** custom trigger state it cannot represent rather than silently dropping it. Schema creation and backfill are independently rerunnable.

## Both consumers, on purpose

The ticket named the scheduler engine. This also moves `brain/systems/cycles/cycle_failure_guard.py` — a second consumer of the shared package, added by #555 hours earlier — onto the same mechanism.

I initially read that as unrequested scope and was ready to cut it, especially since it was the sole source of a regression. The code-quality review argued the opposite and I agree on reflection: the cycle guard runs the identical consecutive state machine, so leaving it behind would make the shared lifecycle abstraction knowingly partial. The regression was caused by the *copied* store machinery, not by including cycles — and giving each guard dedicated state storage dissolved it at the root.

## Verification

```bash
cd /Users/redamjahed/Documents/UwearDev/.codex-worktrees/illospace-538-failure-guard-trigger-state
/Users/redamjahed/Documents/UwearDev/illospace-project/venv/bin/python -m pytest tests/ -m "not requires_db" -q
/Users/redamjahed/Documents/UwearDev/illospace-project/venv/bin/python -m alembic heads
```

**4360 passed, 14 skipped** — full fast suite, run outside the Codex sandbox.
`alembic heads` → `0047_failure_guard_trigger_state (head)`, exactly one head.

## Acceptance checks

1. A **stateful** third trigger registers through the production registry and needs no engine change — it persists its own state and that state actually reaches its evaluation.
2. The existing consecutive and rolling-window triggers behave identically: their pre-existing assertions pass unchanged.
3. `0047` upgrade → downgrade → upgrade is safe, and a downgrade carrying unrepresentable custom state is rejected up front rather than losing data.
4. The cycle failure guard added by #555 still works — `tests/test_cycles_service.py` is back to 85 passed.

## Reviewed

Two Codex code-quality passes. The first returned RESHAPE with four blocking findings — the sharpest being that the "stateful trigger needs no engine change" guarantee rested on **undeclared duck typing**: both engines discovered `evaluate_with_state` via `getattr`, so a trigger could satisfy the public registry contract, persist state, and have that state silently ignored at evaluation. The new test passed only because its trigger happened to know the hidden convention. That is now one declared, typed contract.

The second pass, scoped to the fix delta, returned **`VERDICT: SHIP`** with no blocking findings.

## Known, not fixed here

`brain/systems/failure_guard/core.py:30` uses an invariant context `TypeVar` in input-only protocols; strict PEP 544 checking will want a contravariant one. It does not weaken runtime dispatch, and there is no strict type checking in CI today — flagging rather than silently carrying it.


redawear:
Rechained: `#560` merged `0047_knowledge_index` off `0046_cycle_failure_guard` while this PR was open, so both migrations claimed 0047 off the same parent — each passes CI alone, and whichever merged second would have left **two alembic heads**. This branch now merges current `main` and renumbers to `0048_failure_guard_trigger_state` chained onto `0047_knowledge_index` (both the `Revises:` docstring line and `down_revision`). Verified after the rechain: `alembic heads` returns a single head, and the full fast suite is 4366 passed / 11 skipped.

PR #562, opened 2026-07-28
feat(soul): Illo writes like a teammate, and nothing else gets to decide that

redawear:
Illo's Slack posts read like a run log, and the team stopped reading them. `#4_software` is now **61 Illo messages out of the last 65** — the four human ones are all Reda's.

## Before / after

Both of these are real messages Illo posted on 2026-07-27. The "after" is what the new soul produces, generated through Illo's **real assembled system prompt** on **gpt-5.6-sol, the model Illo actually runs on**:

| | |
|---|---|
| **before** | `Promotion readiness is *BLOCKED* for `staging 1e63153 → main 31d486f`: new blocker `g4:promotion-pr-1296-dirty`. • PR #1296 is open but `mergeable=false` / `dirty`. • `test`, `verify`, and `gitleaks` checks are failing. • Containment and Alembic graph checks did not finish before the hard ceiling…` |
| **after** | Production promotion is blocked: uwear-backend#1296 has a merge conflict, and test, verify, and gitleaks are failing. Containment and Alembic still need checking. |

| | |
|---|---|
| **before** | `@axel-havard new prod issue: `uwear-backend#1374` — media ingestion rejects oversized decoded images. Tracker row filed; coordinator handoff blocked because cycle 2 was already active.` |
| **after** | New production bug: media ingestion rejects images that decode above the size limit. Filed uwear-backend#1374 — Axel, this is with backend. |

## Why the old soul wasn't enough

It was not missing the rule. It already carried Orwell's *"never use a jargon word if you can think of an everyday English equivalent"* and *"write visible replies for a busy human, not for a run log."* The rule was **losing**, for two reasons: nothing named the specific vocabulary that actually leaks, and three code paths let other prose decide the voice.

## What changed

**The soul** (`brain/systems/personality/soul.py`) now:
- names what stays inside — fingerprints, run/cycle ids, commit-pair notation, ledger and handoff bookkeeping, evidence-health labels, budget ceilings
- says lead with the point, say less than you know, 1–4 sentences, let people ask follow-ups
- keeps the link when trimming, treats a digest as a short story not a table dump
- treats silence as a complete outcome — except for a problem the channel already watched arrive
- matches the room's language
- **reserves voice to itself**: missions and guidance say *what* to communicate, never *how* it reads. This is the rule that matters most, because Illo authors its own cycles.

**Three paths that bypassed the soul:**

1. `cycles/prompts.py` declared the mission *"authoritative"* full stop — and the mission sits **last, in the user message**, where it outranks a soul sitting first in the system prompt. Cycle 9's live guidance row #22 exploited exactly this, requiring *"one focus line naming the finding and its fingerprint."* The mission is now authoritative for what is worked on, checked, and reported — **including the result contract's required answer sections** — but never for how it sounds.
2. `cycles/contract_gate.py` rewrites a cycle's visible answer when it misses the contract, using a prompt with **no soul in it**. Every repaired answer shipped in a default register. Fixed.
3. Workers that inherited a response surface wrote user-visible text with **no soul at all**. They get it now via `has_user_visible_surface()`; headless workers stay lean, preserving the original design (there was a deliberate test asserting workers get no soul — it still passes for the headless case).

`SOUL_MAX_CHARS` 6000 → 8000: the default soul is ~6.1k, and `manage_soul` needs to leave Illo room to *edit* its own soul rather than only shrink it. Flagging this explicitly since it bounds operator-authored soul size.

## Verification

- Full suite green: **4361 passed, 85 skipped**.
- Eight recorded Illo messages regenerated through the real `build_fast_system_prompt()` output. Two soul fixes came out of that run rather than from review: **keep the links when trimming** (the first draft dropped `#1169`/`#1175` from a French status reply) and **reply in the language you were addressed in**.
- New tests pin the anti-robot rules and the voice-ownership rule so they cannot silently erode.

## Not in this PR

- **Cycle cleanup.** Cycle 9's guidance #22 still mandates the fingerprint phrasing, and rows #22–24 are marked "RETIRED" in a later row while all three remain `is_active=t` and still reach every run — because `cycle_guidance` is **append-only**: there is no `remove_guidance` action anywhere (`manage_cycle` has `remove_output_target` but not the guidance twin). Retirement is currently expressed by appending more guidance. Needs the primitive, then a cleanup pass.
- `post_slack_reply` still bypasses both the final-reply checker and the contract gate — the least-governed path to a human. Prompt-side only for now.


redawear:
## Review pass — 4 findings, all fixed in a4b04863

Reviewed by a Codex (gpt-5.6-sol) pass and an independent Claude pass. Both landed on the same top finding; the second added one the first missed.

**1. Headless workers would have been given the soul.** `spawn_worker` children inherit the parent's `target_ref` verbatim ([handlers/workers.py:459](https://github.com/Illospace/illospace/blob/main/brain/systems/runs/tool_catalog/handlers/workers.py#L459)), so any headless diagnostic worker under a Slack-origin run carried the parent's `required_response_tool` and read as user-facing. It would then be told to "say less than you know" and keep to "one to four short sentences" — the opposite of what a worker owes its parent (`WORKER_AGENT_INSTRUCTIONS` demands evidence, artifacts, remaining uncertainty). Truncated evidence in exactly the runs gates consume.

**2. `"headless"` is itself a `final_answer_target_surface` value** — monitored intakes, the AWS health pipeline, `work_intake_slack`. The one string meaning "no person reads this" was being read as a person reading it.

**3. The chat path was still bypassed.** Native chat puts `required_response_tool` in **metadata only**, never `target_ref` ([internal.py:316](https://github.com/Illospace/illospace/blob/main/brain/app/triggers/adapters/internal.py#L316)). So chat workers got no soul — the bypass stayed open on the surface people actually use. My original test only passed because it hand-planted the key in `target_ref`; production never does that.

Fix for all three is the same: lead with the headless flag (which `spawn_worker` always sets and `visibility.py` already treats as authoritative), treat the `"headless"` sentinel as what it says, and read surface keys from metadata too. The test now asserts the six real intake shapes instead of a convenient dict.

**4. Soul vs. contract, in the repair path.** The soul named evidence-health labels as internal machinery, while the cycle contract requires an evidence-health section — and the repair prompt now leads with that soul plus "write it in the voice the Soul describes." The one bounded repair pass could have been coached into dropping the section it exists to restore, sending settlement to `mission_contract_failed`. One sentence in "What Stays Inside" now scopes the rule to what you say to people.

**Two corrections to claims in the PR description:** the gate is keyword-loose, not literal-label — `Evidence reviewed:` verbatim is not required, any of evidence/source/reviewed/workspace passes. And `SOUL_MAX_CHARS` 6000→8000 is required rather than convenience: the new default is 6,385 chars and would fail the old limit, which would have poisoned `manage_soul reset`.

**Checked and clean:** no circular import from `brain.systems.cycles` → `brain.systems.personality`; `soul_prompt_section()`'s disk read is not in a hot path; no non-test code references the old soul strings. Confirmed on illo-dev that no `SOUL.md` override file exists at `/data/private/agent-context/`, so the new default is what actually loads — this change is not a no-op in prod.

Suite green after the fixes: 4361 passed, 85 skipped.

PR #564, opened 2026-07-28
fix(deploy): a restart-policy signal must terminate the deploy, and the restore must actually happen

### LINKED ISSUE #552: Worker restart-policy suspension: signal traps do not terminate the deploy
Split out of #544 review. Small, and pre-existing since #527.

## Problem

`suspend_worker_restart_policy` in `deploy/scripts/compose-runtime-lib.sh` installs:

```
trap 'restore_worker_restart_policy' INT TERM HUP
```

The handler restores the policy and returns — then bash carries on. So Ctrl-C or an SSH drop during a worker swap does **not** stop the deploy; it just re-arms the outgoing worker's restart policy and keeps going, potentially resurrecting it beside a live handoff worker. The handler should restore and then terminate with the conventional signal status.

The test that looks like it covers this (`test_worker_restart_policy_suspension_is_restored_when_a_deploy_is_interrupted`) runs `exit 130` rather than sending a signal, so it exercises the EXIT trap and leaves the signal behaviour untested. Worth fixing the test to actually deliver SIGINT.

## Related, and explicitly not fixable here

The suspension is a persistent container mutation (`docker update --restart=no`), so no shell trap covers power loss or SIGKILL — the comment above the function claims the trap "makes the window crash-safe", which overstates it. `reconcile_worker_restart_policy` already repairs the drift on the next deploy, and the boot unit from #527/#529 brings the stack up regardless, so the exposure is bounded. Worth softening the comment to say what the trap actually guarantees.

Also best-effort where they are safety preconditions, from the same review: `docker update --restart=no` is ignored on failure (if it fails, SIGTERM can let Docker resurrect the outgoing worker), and the idle-path kills and handoff removal are unchecked. Verifying them is cheap.

redawear:
Closes #552

`suspend_worker_restart_policy` installed `trap 'restore_worker_restart_policy' INT TERM HUP`, but
the handler restored the policy and **returned** — bash then carried on with the deploy. A Ctrl-C or
an SSH drop mid-swap did not stop anything; it re-armed the outgoing worker's restart policy and
kept going, so Docker could resurrect that worker beside a live handoff worker. Two workers claiming
the same AgentRuns is the exact failure the swap exists to prevent.

## What changed

- **The signal handler terminates.** It resets its own disposition, restores, then re-raises the
  signal against the shell, so the deploy dies with the conventional status (130 / 143 / 129) and
  does no further work.
- **The restore is at-least-once.** It ignores INT/TERM/HUP across the `docker update` and clears
  the suspended-worker ID **only on success**, so an interrupted or failed restore leaves the ID set
  for the EXIT trap to retry. Previously the ID was cleared first, so a second signal mid-restore
  meant the policy was never restored and the EXIT trap could not tell.
- **EXIT-trap ownership is explicit.** Whether we installed the restore is recorded in a state
  variable instead of pattern-matching the caller's trap text — a caller trap that merely mentions
  the function name is no longer mistaken for ours.
- **A failed `docker update --restart=no` aborts** instead of being swallowed, so no caller proceeds
  to `docker kill` believing suspension succeeded. Applied consistently at all three call sites.
- **The comment no longer overstates the guarantee**: the trap covers catchable interruptions, and
  `reconcile_worker_restart_policy` is named as what covers SIGKILL, power loss and persistent drift.
- **The test actually delivers signals.** It previously ran `exit 130`, exercising only the EXIT trap
  and leaving the signal behaviour untested.

## How to check it (no diff reading required)

```
cd /Users/redamjahed/Documents/UwearDev/.codex-worktrees/illospace-552-restart-trap
/Users/redamjahed/Documents/UwearDev/illospace-project/venv/bin/python -m pytest tests/test_safe_deploy.py tests/test_worker_swap_deploy.py -q
```

Acceptance checks, each with a test that fails without the fix:

1. SIGINT/SIGTERM/SIGHUP mid-suspension → policy restored **and** the deploy dies with 130/143/129,
   doing no further work (observed as genuine negative subprocess return codes, not `exit 130`)
2. A second signal *during* the restore → the policy still ends up restored
3. A repeated suspension → the caller's chained EXIT cleanup still runs, with exactly one restore
4. A caller EXIT trap that merely contains the function name → the restore is still installed and runs
5. `docker update --restart=no` failing → surfaced to the caller; no worker gets SIGKILLed on the
   assumption that suspension worked

## Review history

Two Codex code-quality passes, two fix passes.

- **Review 1 → RESHAPE (2 blocking):** the ID was cleared before the restore completed; and a
  repeated suspension clobbered the caller's chained EXIT trap.
- **Review 2 → RESHAPE (1 blocking):** the first fix made the already-installed case a no-op, which
  turned a pre-existing substring false positive into something worse — a caller trap merely
  *mentioning* the function name meant the restore trap was never installed at all. Reproduced by
  the reviewer. Fix pass 2 replaced text matching with explicit ownership tracking.

Review 2 also traced and **cleared** the parts I was most suspicious of: the
`trap -` / `eval "$(trap -p …)"` critical section is mechanically correct including the empty-output
and explicit-ignore cases, and swallowing a second signal during the restore cannot turn a
cancellation into a continued deploy — restoration is only reachable from a signal handler (which
re-raises) or from EXIT (already exiting).

Non-blocking, deliberately left: `restore_worker_restart_policy` returns 0 even when `docker update`
fails — silent, but no caller consumes the status, and returning non-zero would abort the
signal-handler chain before the signal is re-raised. The retained ID means the EXIT trap retries
anyway. The final fix delta (14 production lines) was reviewed by the orchestrator rather than by a
third Codex pass.


redawear:
**Conflict repaired — this is mergeable again.**

Merged current `main` (`560f59f1`, after your seven merges at 00:23–00:25Z) into the branch. One real conflict, in `tests/test_safe_deploy.py`: main added `from brain.contracts import worker_lifecycle as worker_lifecycle_contract` for its new lifecycle assertions, this branch had not. Kept main's import — it is used 10× in the merged file. `deploy/scripts/compose-runtime-lib.sh` auto-merged; both sides' intent is intact.

Verified on the merged tree: **4448 passed, 85 skipped, 0 failed** (full suite).

Re-run it yourself:
```
cd /Users/redamjahed/Documents/UwearDev/.codex-worktrees/illospace-552-restart-trap && ../../illospace-project/venv/bin/python -m pytest tests/ -q
```

<sub>Routine R3 — conflict repair only, no behaviour change to the fix under review.</sub>

Issue #573, opened 2026-07-28
[runs] agent_runs lock ordering is not globally owned — two callers still lock child-before-root


redawear:
Follow-up from the fix for #567, deliberately **not** folded into it: the two call
sites below live in files that open PR #561 is rewriting, so fixing them there would
have created a merge conflict for no gain in the meantime.

## The finding

#567's fix imposes a single lock order on `agent_runs` — ascending run id, rows before
advisory locks — but that order is only enforced for locks acquired **through the
store's helper**. It cannot repair a lock the transaction already holds.

Two callers still take a run row `FOR UPDATE` directly and *then* call
`append_event`, which now requests `FOR KEY SHARE` on the root:

- `brain/systems/runs/cortex/runner.py:897` — `session.get(AgentRunRow, run.parent_run_id, with_for_update=True)`, then `store.append_event(run_event(int(parent.id), "run.worker_failed", ...))` at ~:941.
- `brain/systems/runs/chantier_continuation.py:66` — `anchor = await _lock_run(session, anchor_id)`, then `store.append_event(run_event(anchor.id, ...))` at ~:203.

For a **nested** run — one whose parent/anchor is itself below a higher root — that
acquires the lower-id root *after* the higher-id child, which is exactly the
child-before-root ordering #567 exists to prohibit. The cycle #567 observed in
production (sibling event appends against the liveness path) is closed; this one is
not.

## Why it is an abstraction problem, not just a bug

The store is not the sole owner of `AgentRunRow` locking, so the ordering rule is a
convention that callers can silently violate — and `store.py` is already 1,349 → 1,482
lines. The lock policy wants to be a focused component (e.g. `run_locking.py`) that
owns an explicit root-aware `lock_run_for_update` / transaction context, with direct
`FOR UPDATE` on run rows prohibited outside it.

## Sequencing

**Do this after #561 merges** — it owns both files today.

## Acceptance

- No `FOR UPDATE` / `with_for_update=True` on `AgentRunRow` outside the lock owner.
- The two call sites above acquire through the root-aware helper, so a nested run's
  root is locked before its child.
- A test that fails on the pre-fix ordering: a caller already holding a child lock
  then appending an event must not acquire the root afterwards.

<sub>Filed by Routine R3 from the Codex code-quality review of the #567 fix (BLOCKING finding 1), verified against the call sites before filing.</sub>

redawear:
> *This was generated by AI during triage.*

[...]

PR #575, opened 2026-07-28
fix(runs): order sibling agent_runs locks and retry deadlocks

### LINKED ISSUE #567: [runs] Concurrent spawned sub-runs deadlock on agent_runs row locks and die mid-execution

## What happens

Parallel sub-runs spawned from the same parent deadlock against each other on `agent_runs` row locks. Postgres kills one, and the run fails mid-execution with no domain-level explanation.

Observed on illo-dev, 2026-07-28 (found while verifying the knowledge-index deploy, #560 — unrelated to that change; these failures predate it by hours):

```
run_execution_failed: DBAPIError: (sqlalchemy.dialects.postgresql.asyncpg.Error)
<class 'asyncpg.exceptions.DeadlockDetectedError'>: deadlock detected
DETAIL:  Process 595117 waits for ShareLock on transaction 1235839; blocked by process 595116.
         Process 595116 waits for ShareLock on transaction 1235840; blocked by process 595117.
[SQL: SELECT agent_runs.id, agent_runs.trace_id ...
```

Two processes, each holding a transaction the other is waiting on — a lock-ordering deadlock, not lock contention that would resolve on its own.

## Evidence it is sibling-vs-sibling

Runs `3065` and `3067` share parent `3024` and started at the **same millisecond**:

```
3065 | failed | attempt=0 | started=2026-07-28 18:50:57.745901+00
3067 | failed | attempt=0 | started=2026-07-28 18:50:57.745468+00
```

Both reached `run.activity: "Project context ready"` before dying, so they fail *after* doing real work.

## Scale

- 6 runs hit `asyncpg` errors in 4 hours (deadlock plus at least one bare `TimeoutError`, likely the same contention presenting differently).
- 36 `cortex_idea` runs failed in 3 hours overall.
- `metadata` carries no `failure_reason` on these rows — the cause is only visible in `agent_run_events`, which is why this reads as a silent failure from the run table. Surfacing the reason onto the run would have made this diagnosable in one query instead of four.

## Why it matters

Work is silently lost. Run 3069 retried (`attempt=1`) and failed again 42 minutes later, so the retry path is not reliably rescuing these. Any coordinator fanning out parallel workers is exposed.

## Suggested direction

- Find the `SELECT ... FOR UPDATE` on `agent_runs` in the sub-run start/claim path and establish a deterministic lock order (or take a single lock covering the sibling set) so two siblings cannot acquire in opposite orders.
- Deadlock is retryable by nature — a bounded retry on `DeadlockDetectedError` at the transaction boundary would convert these into a brief delay rather than a dead run.
- Independently: propagate the failure cause onto `agent_runs` so `status='failed'` rows say *why* without a join to events.

Related lock-ordering/identity-map hazard in the same area: #543.

/cc @redawear

redawear:
## Read this first — what this does NOT fix

The lock order this PR introduces is enforced **only for locks taken through the
store's helper**. It cannot repair a lock a transaction already holds, and two callers
outside the store still take a run row `FOR UPDATE` and then append an event:
`cortex/runner.py:897` and `chantier_continuation.py:66`. For a **nested** run those
still acquire child-before-root — the very ordering this PR prohibits.

That is filed as **#573**, sequenced to land after **#561**, which owns both files
today. Editing them here would have created a merge conflict with #561 for no benefit
before it merges. The cycle actually observed in production is closed; that second one
is not.

## The cycle this closes

Verified, not assumed: a child transaction held its own `agent_runs` row while an
event's foreign key requested `FOR KEY SHARE` on the root row, while the liveness path
held the root `FOR UPDATE` and waited on the child. Each waited on a row the other
held. Two sibling sub-runs of parent `3024` (`3065`, `3067`, started in the same
millisecond) died this way after reaching `run.activity: "Project context ready"` —
real work, then a silent death: 6 asyncpg errors in 4h, 36 `cortex_idea` runs failed
in 3h, and the `attempt=1` retry did not rescue them.

## What changed — `brain/systems/runs/store.py` only

- **One global lock order**: every `agent_runs` acquisition goes through
  `_acquire_agent_run_locks`, ascending by run id, row locks before advisory locks.
  The root is taken `FOR KEY SHARE` and only the current row `FOR UPDATE`, so siblings
  no longer contend exclusively on their shared root.
- **Bounded retry on SQLSTATE `40P01`** at the execution-claim transaction boundary,
  logged as `agent_run_deadlock_retry`. Postgres aborts the whole transaction on
  deadlock, so nothing committed is re-run. Cancellation/shutdown keep a separate
  cleanup-and-reraise path.
- **Lock-only calls issue no query off Postgres**, where they could acquire nothing
  anyway.
- `claim_next` acquires once via `_set_status_on_locked_run` instead of locking, then
  re-locking through `set_status`; it keeps a `status != QUEUED` recheck now that
  candidate discovery is an unlocked select.
- `heartbeat_run` preflights status and throttle **before** taking locks, then
  rechecks under them — a throttled or terminal heartbeat no longer takes locks just
  to return `False`. `update_metadata` stays locked: it is always read-modify-write.

No migration. No change to the two files above. Failure-cause propagation (the issue's
third bullet) is untouched — it lives in `cortex/runner.py`, owned by #561.

## How to check it

```bash
venv/bin/python -m pytest tests/ -m "not requires_db" -q
```

**4377 passed, 14 skipped, 0 failed** — run outside the Codex sandbox, on the final
commit. Merges clean against all 8 other open PRs (test-merged pairwise), so this adds
no merge-order constraint.

### Acceptance checks
1. Two sibling sub-runs sharing a parent acquire `{root, self}` in ascending id order,
   so the observed cycle is unreachable — not merely less likely.
2. A `40P01` at the claim boundary retries a bounded number of times and says so in the
   logs; it never fails silently.
3. SQLite-backed tests still pass and issue no lock query — every lock site stays
   dialect-guarded.
4. A throttled heartbeat takes no row lock.

### Assumptions worth challenging
- The observed production cycle is the store-owned one (event-append FK vs liveness).
  If the deadlocks continue after this lands, #573 is the next suspect, not a
  regression here.
- The retry bound is 3 with exponential backoff from 50ms — tuned by reasoning, not by
  measurement against production contention.

Closes #567


redawear:
**R3 repair (2026-07-29) — conflict resolved, this is mergeable again.**

#580 rewrote `brain/systems/runs/store.py`, the single file this PR changes, which left it CONFLICTING. Repaired by merging current `main` (`f24f4cae`) into the branch — one conflict, in `create_run`:

- #580 added `admitted_at = datetime.now(timezone.utc)` (feeds `started_at` and `deadline_at`).
- #567 added the parent/root ordered lock acquisition.

**Both are kept**, lock first and the timestamp second, so the admission stamp reflects the moment the run is actually admitted rather than the moment before it waited on the lock.

Fast suite on the merged branch: **4458 passed, 14 skipped, 0 failed**.

PR #681, opened 2026-08-05
fix(runs): make sibling runs on one thread visible to each other (#671)

### LINKED ISSUE #671: [runs][slack] Sibling runs on one thread are mutually blind — a shallow answer raced a mid-flight investigation of the same question

## Problem

Two runs can serve the same Slack thread at the same time without knowing about each other. One run can post a shallow answer while a sibling run is mid-investigation on the same question. The human sees one agent that contradicts itself minutes apart.

## Evidence (illo-dev, 2026-08-04, support request #6)

- Run 14644 (intake, 18:51:53–18:55:48Z) was investigating the customer's generation payloads: 11 `search_knowledge` calls, 5 prod-Postgres `run_script` calls, filed uwear-backend#1571.
- Run 14645 (18:54:01–18:54:18Z) answered Reda's "which image model is she trying to use?" on the same thread, concurrently, from the report text alone: "She doesn't specify the image model in the report."
- Three minutes later 14644's investigation surfaced the exact answer (GPT Image 2 high via fal, five rejected attempts, refunds). Reda had already been given the shallow answer and had to push back.

A human teammate in that state says "checking the payloads now — one sec." Illo said "the report doesn't say."

## Fix direction (open)

Options, not prescriptions:
- Give a reply run visibility of active sibling runs on the same anchor thread (they share `thread_id`/anchor), so it can defer, join, or reference the in-flight work.
- Or serialize interactive replies per thread behind the in-flight investigation when one exists.

## Acceptance criteria

1. When a question arrives on a thread with an active investigating run, the reply either waits for it, references it, or explicitly hands off — it does not answer from surface text as if no investigation exists.

redawear:
Closes #671

> *This PR was built by an autonomous routine (R3): Codex implemented, Claude directed and reviewed. It is a draft — Reda merges.*

> **⚠️ Stacked on [PR #676](https://github.com/Illospace/illospace/pull/676) — merge that one first.** This PR targets `illo-qa/667-open-ask-terminal-states`, not `main`, because both change `tool_catalog/handlers/slack.py` and they genuinely conflict (`git merge-tree` reports a content conflict there). GitHub will retarget this to `main` automatically once #676 merges. The diff below shows only this ticket's changes.

## What this fixes

Two runs can serve the same Slack thread at the same time without knowing about each other. On 2026-08-04, run 14644 was 11 `search_knowledge` calls and 5 prod-Postgres queries deep into the customer's generation payloads when run 14645 answered the same thread from the report text alone: *"She doesn't specify the image model in the report."* Three minutes later 14644 had the exact answer. You had already been given the shallow one.

A human teammate in that state says "checking the payloads now — one sec."

## How

Two halves. Visibility alone would not have prevented this — a model that is merely *told* about a sibling can still answer from surface text.

1. **Visibility.** Active same-thread runs are now attached at admission and rendered in the run's context as authoritative state. This generalises the query that already existed for status questions (`status_questions.py`) rather than adding a second concurrency system — no locks, no queue, no serialisation.
2. **Enforcement.** A reply that ignores a live sibling is rejected. The run must **declare** how it is coordinating — `{"action": "wait"}`, `{"action": "reference", "run_id": N}`, or `{"action": "handoff"}`.

## What to check

1. A question arriving on a thread with an active investigating run gets a reply that waits, references the sibling, or hands off — never a surface-text answer as if nothing were running.
2. A thread with **no** active sibling behaves exactly as before. This is the regression that matters most: the rule runs on every interactive reply.
3. After deploy, the next real concurrent thread is the live test. Watch for a reply that says it is waiting when nothing is actually running — that is the conservative-deferral trade below.

## Three things worth knowing before you merge

- **Coordination is a declared value, not a guess at English.** The first implementation matched phrases in the reply body. The maintainability review broke it in both directions: *"I'll let the investigation already under way finish"* was blocked, while *"one moment — the other run is irrelevant; she did not specify the model"* passed. Deterministic code cannot judge a sentence, so it no longer tries; it verifies a typed declaration instead.
- **The block is bounded, per persisted run.** After two blocked replies the third is admitted as advisory and reaches you. A gate that could block every retry would be strictly worse than the bug this fixes — a shallow answer at least arrives. The first version of the bound reset on every execution attempt, so a requeued run got two fresh blocks; the count is now persisted in run metadata and rehydrated per attempt, with a test that blocks twice, re-executes the same run, and asserts the third reply comes through.
- **Any open same-thread run counts as active work — deliberately.** There is no typed "investigating" state in the schema, so a narrower test would be guesswork. The cost is conservative deferral for unrelated same-thread work. The rationale is written down at the policy, not left implicit.

## Verification

- **Host fast lane, the repo's own CI command, whole suite:** `4757 passed, 11 skipped, 83 deselected` on `main`, and `4762 passed, 11 skipped, 84 deselected` after the rebase onto #676's branch. Zero failures both times. Baseline on `main` is 4750; the extra 7 are this ticket's tests, and the remaining 5 are #676's own.
- Tests cover the acceptance case (context names the sibling; the shallow answer is blocked; a declared wait is accepted), the no-sibling regression, the legitimate paraphrase that must pass, the deceptive phrase that must fail, and the bound.
- No schema change, no alembic migration.
- Two Codex maintainability reviews. The first returned three blocking findings, the second confirmed two structurally resolved and caught the reset-on-restart bound. Both folded. One honest negative it volunteered and I am passing on unchanged: **conditional count is 658 before and 658 after** across the 16 touched files — reply policy moved into a better owner (a shared `outbound_reply_admission` service both the Cortex and Slack handlers call, with `tool_execution.py` back to its original 989 lines and no reply-tool names in it), but complexity relocated rather than left.

## Design note

The fix direction was left open on the ticket. A [read-only investigation](https://github.com/Illospace/illospace/issues/671#issuecomment-5185202034) costed three options against the real code first — sibling visibility, per-thread serialisation, and steering into the active run — and picked this one. Serialisation is survivable but buys a convoy, because "investigating" is not a typed state; steering races the active loop's final drain.

### DIFF
--- a/brain/systems/runs/context.py
+++ b/brain/systems/runs/context.py
@@ -10,6 +10,7 @@ from brain.systems.runs.skill_commands import parse_slash_skill_names
 from brain.systems.runs.status_questions import (
     format_active_sibling_context,
     format_status_question_context,
+    is_status_question,
 )
 
 PROMPT_REFERENCE_CHAR_LIMIT = 12_000
@@ -196,8 +197,7 @@ class RunContext:
     skills: list[str] = field(default_factory=list)
     handoff: dict[str, Any] = field(default_factory=dict)
     request_source: dict[str, Any] = field(default_factory=dict)
-    active_sibling_context: dict[str, Any] = field(default_factory=dict)
-    status_question_context: dict[str, Any] = field(default_factory=dict)
+    same_thread_run_context: dict[str, Any] = field(default_factory=dict)
     scheduled_result_contract: bool = False
 
     def prompt_context(self) -> str:
@@ -246,12 +246,14 @@ class RunContext:
             )
         if self.memory:
             parts.append("Memory:\n" + "\n".join(f"- {item}" for item in self.memory))
-        sibling_context = format_active_sibling_context(self.active_sibling_context)
-        if sibling_context:
-            parts.append(sibling_context)
-        status_context = format_status_question_context(self.status_question_context)
-        if status_context:
-            parts.append(status_context)
+        if is_status_question(self.message):
+            status_context = format_status_question_context(self.same_thread_run_context)
+            if status_context:
+                parts.append(status_context)
+        else:
+            sibling_context = format_active_sibling_context(self.same_thread_run_context)
+            if sibling_context:
+                parts.append(sibling_context)
         return "\n\n".join(parts)
 
 
@@ -274,14 +276,9 @@ class RunContextLoader:
             workspace_ref=dict(workspace_ref or {}),
             thread_context=dict(thread_context) if isinstance(thread_context, dict) else {},
             request_source=_request_source_from_metadata(metadata),
-            active_sibling_context=(
-                dict(metadata.get("active_sibling_context") or {})
-                if isinstance(metadata.get("active_sibling_context"), dict)
-                else {}
-            ),
-            status_question_context=(
-                dict(metadata.get("status_question_context") or {})
-                if isinstance(metadata.get("status_question_context"), dict)
+            same_thread_run_context=(
+                dict(metadata.get("same_thread_run_context") or {})
+                if isinstance(metadata.get("same_thread_run_context"), dict)
                 else {}
             ),
             skills=_skill_names_from_metadata(metadata, message),
--- a/brain/systems/runs/context.py
+++ b/brain/systems/runs/context.py
@@ -7,7 +7,10 @@ import json
 from typing import Any
 
 from brain.systems.runs.skill_commands import parse_slash_skill_names
-from brain.systems.runs.status_questions import format_status_question_context
+from brain.systems.runs.status_questions import (
+    format_active_sibling_context,
+    format_status_question_context,
+)
 
 PROMPT_REFERENCE_CHAR_LIMIT = 12_000
 PROMPT_SCALAR_CHAR_LIMIT = 1_200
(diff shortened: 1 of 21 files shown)


redawear:
> *This was generated by AI during triage (Routine R3).*

**This PR has no CI, and that is not visible from the checks section.** Its base is
`illo-qa/667-open-ask-terminal-states` rather than `main`, and all three PR workflows
filter on `pull_request: branches: [main]`, so nothing ever triggered:

```
$ gh run list --branch illo-qa/671-sibling-run-visibility
(no output — zero workflow runs have ever existed for this branch)
```

The empty checks list reads as "no checks configured", not "never tested". Filed as
#682 and fixed in #683; that fix cannot retroactively cover this PR, because GitHub
reads workflow files from the PR's merge ref and this branch's parent predates it.

**So I ran the repo's own CI command by hand instead**, on this branch's head
`d38dca67`:

```
$ python -m pytest -m "not requires_db and not requires_browser and not live_provider" -q
4762 passed, 11 skipped, 84 deselected in 73.86s
```

That is the same command and marker expression `Backend fast suite` runs. Green.

**Still not covered:** the `Backend DB suite` lane (`-m requires_db`, part of the 84
deselected above) needs Postgres and did not run anywhere — not in CI, not locally.
Every sibling PR in this batch got that lane green in GitHub; this one has no result
either way.

redawear:
Follow-up: the DB lane is covered now too. I brought up pgvector/pg16 on 5433 — the same image and port `Backend DB suite` uses — ran `alembic upgrade head` (to `0051_open_ask_terminal_states`, from this branch's parent) and then the lane itself:

```
$ python -m pytest tests/ -m requires_db -q
80 passed, 4746 deselected in 20.77s
```

So both CI lanes are green on `d38dca67`, by hand rather than by GitHub. Nothing on this PR is unverified any more.

Issue #682, opened 2026-08-05
[ci] PRs based on a feature branch run no CI at all — stacked PRs merge untested


redawear:
> *This was generated by AI during triage.*

## Problem

**A pull request whose base is not `main` never runs any CI in this repo.** All three
PR-triggered workflows filter the base branch:

- `.github/workflows/brain-ci.yml:4-5` — `pull_request: branches: [main]`
- `.github/workflows/deployment-smoke.yml:4-5` — same
- `.github/workflows/container-images.yml:7-8` — same

So a stacked PR — one deliberately based on another feature branch — is merged on the
strength of no test run at all, and nothing on the PR page says so. The checks section
is simply empty, which reads like "no checks configured", not "this code was never
tested".

## Evidence

PR #681 (issue #671) is based on `illo-qa/667-open-ask-terminal-states` rather than
`main`:

```
$ gh pr view 681 --json baseRefName,statusCheckRollup
{"baseRefName":"illo-qa/667-open-ask-terminal-states","statusCheckRollup":[]}

$ gh run list --branch illo-qa/671-sibling-run-visibility
(no output — zero workflow runs have ever existed for this branch)
```

Its parent branch, based on `main`, got the full set on the same day:

```
$ gh run list --branch illo-qa/667-open-ask-terminal-states
completed  success  Brain CI          pull_request  5m29s  2026-08-04T22:01:55Z
completed  success  Deployment Smoke  pull_request  4m22s  2026-08-04T22:01:54Z
```

The fast suite passes on #681 — but only because it was run by hand on a laptop. That
is not a control.

This is newly load-bearing: stacked pull requests became the standard way to ship
dependent work here, so the number of PRs that silently skip CI goes up from now on,
not down.

## Fix

Drop the `branches: [main]` filter from the `pull_request:` trigger in all three
workflows, keeping every `paths:` filter as-is. Any PR then runs the checks its
changed paths select, regardless of base.

Safe to broaden: `container-images.yml` already builds without publishing on PR
events (`push: ${{ github.event_name != 'pull_request' }}` at lines 129, 177, 225),
so a wider trigger builds more images but pushes none.

Note the one limit this cannot fix: for `pull_request` events GitHub reads workflow
files from the PR's merge ref, so a child branch only picks the change up once its
parent contains it. Stacked PRs opened before this lands still need a local run.

## Acceptance checks

1. Open a PR whose base is a feature branch; Brain CI runs on it and reports on the
   PR page.
[...]

PR #695, opened 2026-08-05
fix(knowledge): never index a memory node that no org can reach (#680)

### LINKED ISSUE #680: [knowledge] Org-less memory nodes are indexed but unsearchable — null org_id matches no org

Found while building the skills connector (#674, PR #679). Out of scope there because it is a behavior change, not a refactor — filing it so the decision is made deliberately.

## Problem

Memory nodes with no `org_id` are mirrored into `knowledge_items` with `extra["org_id"] = None`, and the search filter admits an item only when its org matches the caller's org. `None` matches no org, so **those items are permanently unsearchable** — they are indexed, embedded, and unreachable.

- `MemoryNode.org_id` is nullable (`brain/platform/db/models/reconstructive_memory.py:81`), so org-less nodes are a normal state, not corruption.
- The connector writes the null through at `brain/systems/knowledge/connectors/memory.py:72` and `:102`.
- Before #674 the filter was `extra["org_id"] == org_id` with no other branch, so there was no way for such an item to be returned.

This is invisible from the outside: sync reports them as ingested, and search simply never returns them. Nobody sees an error.

## What #674 changed, and what it deliberately did not

PR #679 replaced the org filter with an explicit scope contract: `KnowledgeDraft.scope` is a required `organization | global` field, persisted as `extra["knowledge_scope"]`, and search admits an item when the org matches **or** the item declares global scope.

That gives this bug a clean fix — an org-less memory node can now declare `global` instead of carrying a null org. #679 deliberately did **not** apply it: the memory connector still declares `organization`, so today's behavior is preserved exactly. Making these items visible changes what `search_knowledge` returns, and that is a decision, not a cleanup.

## The decision this ticket needs

Which is true of an org-less memory node?

1. **It is global** — knowledge that belongs to the workspace rather than one org. Then the memory connector declares `KnowledgeScope.GLOBAL` for `org_id is None` and they become searchable.
2. **Its org is merely unknown** — a data-quality gap. Then the fix is upstream (backfill the org, or stop creating org-less nodes) and they stay unsearchable until the org is known.

Do not answer it by inference from the code — the two cases look identical there. Check what actually creates org-less nodes and how many exist.

## Acceptance criteria

1. A written answer to the question above, in the ticket or in `specs/knowledge-base/`.
2. Whichever branch is chosen, an org-less memory node is either retrievable by `search_knowledge` or provably not created any more — no third state where it is indexed and unreachable.
3. A test that pins the chosen behavior, so the next scope change cannot silently flip it.

redawear:
Closes #680.

> *This was generated by AI. A human should review it before merge.*

## What this fixes

A memory node with no `org_id` was mirrored into `knowledge_items` with
`extra["org_id"] = None`. Search admits an item only when its org matches the
caller's, and `None` matches no org — so the item was **indexed, embedded, and
permanently unreachable**. Nothing surfaced it: sync reported it ingested, and
search simply never returned it.

**Zero such nodes exist today** — 0 of 938, counted on the live database. So this
is not a repair, it is closing the door: the state was *absent from the data* but
still *representable by the code*.

## The decision this ticket asked for, answered from the database

The ticket asked whether an org-less node is (1) global knowledge or (2) a
data-quality gap, and said explicitly not to answer it by reading code. Counted
on `illospace-postgres-1`:

| org-less | with org | total |
|---:|---:|---:|
| **0** | 938 | 938 |

Answer: **(2), a data-quality gap.** Option 1 would pin a semantic that nothing
exercises, and in a single-org workspace `organization` and `global` scope are
observationally identical anyway. Full reasoning in
[this comment](https://github.com/Illospace/illospace/issues/680#issuecomment-5192417444).

So the memory connector keeps declaring `KnowledgeScope.ORGANIZATION` — **this PR
changes nothing about what the 938 real nodes do.**

## What changed

One production file, `brain/systems/knowledge/connectors/memory.py`:

- A shared node with no `org_id` is **not mirrored**. It is skipped with a
  warning naming the node id.
- `_required_org_id()` raises at the draft-construction boundary, so a null org
  cannot reach an active knowledge item even if the selection logic is changed
  later.
- Mirroring and withdrawal are now modelled separately (see below).

No migration, no schema change, no change to `KnowledgeScope`, the search filter,
or any other connector. Tightening `memory_nodes.org_id` to `NOT NULL` is a
separate call — the connector guard delivers the safety without one.

## Review caught a leak that the tests could not

The first implementation applied the org guard **before** choosing between
mirroring and withdrawal. A node that was mirrored while it had an org, then went
private *and* lost its `org_id`, was skipped instead of withdrawn — so its stale
knowledge item **stayed visible to its former organization**. That is the same
bug class this ticket exists to kill, reintroduced on the withdrawal path.

Fixed: active mirroring requires an org on the node; **withdrawal never does** —
it takes the org from the existing mirror, which is non-null by construction
because this guard is what prevents a null-org item from being written. There is
a regression test for exactly this transition, and it fails against the first
commit (`b27ed548`) and passes now.

A first attempt also added a `skipped` counter to `KnowledgeEnumeration`, a
dataclass five connectors return and one would set, folded into a `stats.skipped`
that already means "unchanged". That was my instruction and it was wrong — it
conflated healthy with invalid. Reverted; the warning log is the whole reporting
surface.

## Verification

Both lanes run by hand, because **this PR gets no CI** — its base is a feature
branch, which is #682 / #683.

- Fast lane, the repo's own CI command: **4753 passed, 11 skipped, 0 failed.**
  Base branch is 4751 and this adds exactly 2 tests, 0 removed (82 insertions,
  0 deletions under `tests/`).
- DB lane on a clean `pgvector/pgvector:pg16` container, migrated to this
  branch's head `0050`: see the comment below.

## Acceptance checks

1. An org-less memory node is never written to the index — assert on the absence
   of both the `KnowledgeItem` and its embedding, not just a count.
2. A node with a real org is still mirrored and still declares
   `KnowledgeScope.ORGANIZATION` — no behaviour change for the 938 real nodes.
3. A previously mirrored node that goes private **and** org-less is still
   withdrawn, so its existing item is scrubbed rather than stranded.
4. No migration and no new alembic revision — deliberate, since revision ids are
   contested by other open PRs.

## Note for the reviewer

`existing_org_ids` reads `KnowledgeItem.extra["org_id"]` directly rather than
`.get()`. That is intentional: this connector is the only writer for its
`source_key` and always sets the key, so a missing key means a genuinely
malformed item and should fail loudly (the enumeration error is caught and
logged by `sync_connector`) rather than silently withdraw with a null org.

## Stacking

Based on `illo-qa/674-skills-knowledge-connector` (#679), not `main`. The
`KnowledgeScope` contract this builds on exists only there, and `memory.py` is
owned by that PR — a `main`-based branch would collide. **Merge #679 first.**

### DIFF
--- a/brain/systems/knowledge/connectors/base.py
+++ b/brain/systems/knowledge/connectors/base.py
@@ -115,6 +115,7 @@ class KnowledgeEnumeration:
     drafts: list[KnowledgeDraft]
     cursor: dict[str, Any]
     failures: tuple[EnumerationFailure, ...] = ()
+    skipped: int = 0
 
 
 class KnowledgeConnector(Protocol):
--- a/brain/systems/knowledge/connectors/base.py
+++ b/brain/systems/knowledge/connectors/base.py
@@ -115,7 +115,6 @@ class KnowledgeEnumeration:
     drafts: list[KnowledgeDraft]
     cursor: dict[str, Any]
     failures: tuple[EnumerationFailure, ...] = ()
-    skipped: int = 0
 
 
 class KnowledgeConnector(Protocol):
--- a/brain/systems/knowledge/connectors/memory.py
+++ b/brain/systems/knowledge/connectors/memory.py
@@ -77,7 +77,7 @@ def _draft_for_memory(
     )
 
 
-def _withdrawn_draft(node: MemoryNode) -> KnowledgeDraft:
+def _withdrawn_draft(node: MemoryNode, *, org_id: str) -> KnowledgeDraft:
     """Scrub a formerly shared mirror after its source becomes private."""
 
     return KnowledgeDraft(
@@ -92,7 +92,7 @@ def _withdrawn_draft(node: MemoryNode) -> KnowledgeDraft:
             "archived": True,
             "mirror_status": "visibility_withdrawn",
             "node_kind": node.node_kind,
-            "org_id": _required_org_id(node),
+            "org_id": org_id,
             "truth_status": node.truth_status,
             "visibility": node.visibility,
         },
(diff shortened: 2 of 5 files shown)


redawear:
## DB lane, run by hand

This PR's base is a feature branch, so GitHub runs **no checks on it at all** (that is #682 / [#683](https://github.com/Illospace/illospace/pull/683)). Both lanes were therefore run locally.

Clean `pgvector/pgvector:pg16` container, migrated to this branch's head:

```
INFO  [alembic.runtime.migration] Running upgrade 0049_agent_run_deadlines -> 0050_scheduler_cold_start
===================== 79 passed, 4737 deselected in 19.41s =====================
```

Fast lane (the repo's own CI command, `-m "not requires_db and not requires_browser and not live_provider"`):

```
========== 4753 passed, 11 skipped, 83 deselected in 75.39s (0:01:15) ==========
```

A fresh container was needed because the shared local test DB on port 5433 is stamped at `0051` from #676's branch, so `alembic upgrade head` fails there. This branch's head is `0050` — it adds no revision.

PR #693, opened 2026-08-05
fix(cycles): make the promotion-readiness cheap exit real, and record a verdict every run (#689)

redawear:
Closes #689

> **Stacked on #672.** Base is `illo-qa/668-self-review-persistence`, not `main`. #672 makes `self_review_summary` store the agent's self-review instead of the token bill, and this work is meaningless without it. **Merge #672 first, then this** — or merge the train in one click.

## The finding

Cycle 9 (*Uwear Backend Promotion Readiness*) completed six runs at 2.2M–4.8M tokens each — about **$28.50** — and posted **one** message in its entire lifetime. Its contract promised cheap exits: *"IDLE/UNCHANGED under 1 minute."*

**Those exits existed only in the cycle's mission prompt.** The runtime always admitted a full agent with GitHub review tools, so nothing enforced them. There was no executable checkpoint between the SHA comparison and the per-PR sweep — the cheap path was a sentence, not a gate.

That also explains the silence. The posting gate was agent-controlled and sat *after* the expensive sweep, so a run that burned its budget getting there posted nothing and recorded nothing. A run that evaluated staging and chose silence was indistinguishable from one that never arrived.

## The change

1. **A real gate that runs before the sweep.** IDLE finishes before agent admission. UNCHANGED skips per-PR review, disables the GitHub source tool, and still continues to the posting decision. Changed-and-ahead allows the full path. Failed cheap reads continue with degraded evidence rather than wrongly skipping.
2. **A verdict on every run, including silent ones.** `evaluated_low_silent`, `skipped_unchanged` and `gate_not_reached` are now distinct, each with readable `Risk:` / `Evaluated:` / `Posted:` lines. A run that dies before the gate records `Risk: UNKNOWN` instead of looking like a deliberate silence.

Item 3 of the ticket — *why* runs 2220, 2277 and 2443 stayed silent — is an investigation, and (2) is what makes it answerable next time. The evidence for the budget-exhaustion hypothesis is above; the old records cannot prove it, so no speculative fix was written for it.

## This took two rounds

A maintainability review returned CHANGES REQUIRED with five blocking findings. The behaviour was right; the seams were wrong. All fixed:

1. **The gate keyed off the cycle's display name.** Renaming it in the UI would silently remove the whole protection and restore the burn. A durable key needs a migration (out of scope — see below), so instead the failure is now **loud**: missing, renamed or ambiguous configuration emits a configuration alert and fails the run with a readable terminal verdict. Silently returning `None` was the actual danger.
[...]

Issue #689, opened 2026-08-05
[Illo-QA][cycles] Promotion readiness burns 2.2M–4.8M tokens per run and has posted once in its lifetime


redawear:
> *This was generated by AI during triage.*

## Problem

Cycle 9 (*Uwear Backend Promotion Readiness*) has completed six runs at 2.2M–4.8M tokens each and has posted **one** message in its entire lifetime. Its contract defines cheap exits for the idle case — *"IDLE/UNCHANGED under 1 minute"* — but the silent runs burned millions of tokens, so they were not taking the cheap path. A run that evaluated staging and stayed silent is indistinguishable from a run that skipped, because nothing is recorded either way.

## Evidence

**1. Burn per run**, from `cycle_runs.self_review_summary`:

| run | scheduled | tokens | cost |
|---|---|---:|---:|
| 2506 | 2026-08-04 15:00 | 4,202,708 | $5.68 |
| 2443 | 2026-08-03 15:00 | 4,824,870 | $6.19 |
| 2337 | 2026-08-01 16:40 (off-slot) | 2,497,407 | $3.21 |
| 2277 | 2026-07-31 15:00 | 4,656,085 | $7.46 |
| 2220 | 2026-07-30 15:00 | 3,065,934 | $3.68 |
| 2198 | 2026-07-29 22:08 | 2,154,278 | $2.28 |

≈ **$28.50** across six completed runs.

**2. Output: one message, ever.** A workspace-wide Slack search for the cycle's own headline phrase `"Staging risk today"` returns exactly one result — `2026-08-04 11:07 EDT`, `#4_software`, ts `1785856047.124609`. Runs on 07-30, 07-31 and 08-03 produced no visible output at all.

**3. The posting condition was met throughout.** The contract says to post *"on scheduled weekday runs whenever staging is ahead"*. `main...staging` is currently `uwear-ai/uwear-backend` **47 ahead / 12 behind** and `uwear-ai/uwearaiapp` **93 ahead / 2 behind**, and the 08-03 weekly brief independently reported the promotion PRs blocked — staging has been continuously ahead across the whole period.

[...]

PR #697, opened 2026-08-05
feat(cycles): a durable execution-policy key, resolved through a registry (#694)

### LINKED ISSUE #694: [cycles] A Cycle cannot durably declare its execution policy — the gate keys off a mutable display name

## Problem

A Cycle has no durable way to declare which execution policy applies to it. The only identity available is `Cycle.name` — a mutable UI label — or `Cycle.id`, which is database-specific and not portable across environments.

This surfaced in #689 ([PR #693](https://github.com/Illospace/illospace/pull/693)). The promotion-readiness cheap gate has to find "the Cycle this policy belongs to", and the only handle it has is string equality against the literal `"Uwear Backend Promotion Readiness"`. One mutable string controls gate execution, contract creation and terminal fallback recording.

**Renaming that cycle in the UI removes the entire protection**, and the protection in question is what stops a run burning 2.2M–4.8M tokens (~$5) to post nothing.

#693 mitigated the danger — a missing, renamed or ambiguous match now emits a configuration alert and fails the run with a readable verdict, instead of silently returning `None`. That makes the failure loud. It does not make the identity durable.

## Why it was not fixed in #693

The fix needs a new column on `Cycle` and `CycleRevision`, so it needs an alembic migration. Revision `0051` was already taken by another open PR (#676), and #693 was explicitly scoped to add no migration.

## The second half of the same problem

`async_execute_cycle_run` still imports the promotion-readiness module directly. #693 moved the *knowledge* out — the executor now handles only a generic execution effect (finalize, or admit with a metadata patch) and no longer mentions `unchanged` or `read_github_source`. But the wiring is still hardcoded, so a second cheap gate means a second import and a second branch.

These two halves are one fix: a cycle declares a stable policy key, and a registry resolves that key to a gate. Doing the registry without the durable key just moves the string matching.

## Proposed fix

1. Add a durable `execution_policy_key` (or `pre_admission_gate_key`) to `Cycle` and `CycleRevision`, with a migration.
2. Backfill Cycle 9 with `uwear_backend_promotion_readiness`.
3. Resolve the key through a registry, and reject unknown configured keys at create, update and execution time — an unrecognised key must fail loudly, never silently skip.
4. Register the promotion-readiness gate as the first entry. Consider making `auth_preflight` and `quota_preflight` default stages in the same framework; they already apply to every cycle and return neutral `blocked` / `deferred` results, so they are the natural precedent.
5. Delete the name-matching path and its configuration-alert scaffolding once the key is authoritative.

## Acceptance checks

1. Renaming Cycle 9 in the UI does not disable its gate.
2. A cycle configured with an unknown policy key fails loudly at save time, not silently at run time.
3. Adding a second cheap gate requires no edit to `async_execute_cycle_run`.
4. The promotion-readiness behaviour from #689 is unchanged: IDLE finishes before admission, UNCHANGED skips per-PR review but still reaches the posting decision, and every run records a `Risk:` / `Evaluated:` / `Posted:` verdict.

## Related

- #689 / [PR #693](https://github.com/Illospace/illospace/pull/693) — where this was found and mitigated
- #668 / [PR #672](https://github.com/Illospace/illospace/pull/672) — the persistence surface #693 stacks on
- #676 — holds alembic revision `0051`

redawear:
Closes #694.

## What this does

A Cycle had no durable way to declare its execution policy. The promotion-readiness gate — the thing that stops a run burning 2.2M–4.8M tokens (~$5) to post nothing — found its Cycle by string equality against the display name `"Uwear Backend Promotion Readiness"`. **Renaming that Cycle in the UI removed the protection entirely.** #693 made that failure loud; it did not make the identity durable.

This adds a durable `execution_policy_key` on `Cycle` and `CycleRevision`, resolves it through a registry, and deletes the name-matching path.

**The coupling is gone, not just guarded:** the gate path no longer reads `cycle.name` at all. The literal string now survives only inside the one-time migration backfill and its test.

## Stacked — this is a layer, not a standalone PR

Base is `illo-qa/689-promotion-readiness-verdict` (#693), which is on `illo-qa/668-self-review-persistence` (#672). The `CycleExecutionEffect` contract this builds on does not exist on `main`. Merging the train merges this.

## The migration — read this before merging alongside #676

`0052_cycle_execution_policy_key`, `down_revision = 0050_scheduler_cold_start`.

Revision `0051` is held by **#676**, which is on a different train and is *not* in this branch's ancestry. The number `0052` is deliberate: it avoids a filename collision, and this branch reports a **single alembic head** on its own.

⚠️ **If #676 merges first, this migration's `down_revision` must be re-pointed to `0051_open_ask_terminal_states`** — a one-line edit. Otherwise the two land as parallel heads. Whoever merges second makes that change.

The backfill sets Cycle 9 and its revisions by name (the only identity that exists at migration time) and is a **no-op if that Cycle is absent**, so it is safe on a fresh or non-production database. Both cases are pinned by test.

## Review — one blocking finding was folded

A cross-family maintainability review returned **CHANGES REQUIRED**. The blocking finding was real and worth the round trip:

The first registry (`451cb3b5`) was a mutable list with `register()`/`unregister()`. It accepted a **duplicate key silently** and only reported the ambiguity later, at validation or execution time — turning an import-time configuration error into a failed production run. That is the same late-failure class this ticket exists to remove.

This package already had the right pattern: `CycleFailureGuardRegistry` in `brain/systems/cycles/cycle_failure_guard.py:316` is a frozen dataclass over a tuple validating uniqueness in `__post_init__`, behind a factory. `88717056` rebuilds the policy registry in that shape rather than shipping a second, weaker one. Verified directly:

```
duplicate key -> ValueError at construction: Ambiguous Cycle execution policy key: 'k' resolves to 2 gates
blank key     -> ValueError at construction: Cycle execution policy registration requires a key
```

Two non-blocking findings were also folded: the gate is now a `Protocol` with the exact call signature (so a mis-shaped gate cannot register and fail only when a run calls it), and a dispatcher-level test was added — see acceptance check 1.

**One finding was declined.** The review objected to `tests/test_cycle_execution_policy_migration.py` being listed in `LEGACY_SYNC_TEST_HARNESSES` in `scripts/async_db_debt_metrics.py`. That list already contains `tests/test_cold_start_reconciliation.py`, the migration test for `0050`, so this follows the repo's established precedent for migration tests. Re-classifying that scheme is a separate concern and would widen an already-large diff.

## Verification — run by hand, because this PR gets no CI

**This PR's base is a feature branch, so GitHub runs zero checks on it** — the gap #683 fixes. Every figure below I ran locally with the repo venv, outside the Codex sandbox (which reports phantom `127.0.0.1` bind failures):

```
pytest -m "not requires_db and not requires_browser and not live_provider" -q
  -> 4774 passed, 11 skipped, 83 deselected
     (base cc26c56f = 4766; +8 reconciles exactly against tests added/superseded)

alembic upgrade head        (fresh pgvector/pgvector:pg16, empty database)
  -> 0050_scheduler_cold_start -> 0052_cycle_execution_policy_key, clean
     alembic heads -> ['0052_cycle_execution_policy_key']  (single head)

pytest tests/ -m requires_db -q
  -> 79 passed

python -m brain.jobs.check_cycle_context_admission
  -> {"ok": true, "cycle_count": 3}
```

## Acceptance checks

1. **Renaming Cycle 9 does not disable its gate.** ✅ Proven twice: structurally (the gate path never reads `cycle.name`) and by `test_renamed_cycle_reaches_gate_through_real_execution_policy_dispatcher`, which drives a renamed Cycle to its gate through the **real** dispatcher rather than a patched-out stub.
2. **An unknown policy key fails loudly at save time, not run time.** ✅ `async_create_cycle` raises before adding or flushing; `async_update_cycle` validates the effective stored key before any mutation, even when the patch does not touch the key. REST callers get HTTP 400. At execution time an unresolvable key finalizes the run as `failed` with the reason in `context_snapshot`.
3. **A second cheap gate needs no edit to `async_execute_cycle_run`.** ✅ `test_execute_cycle_run_applies_a_second_registered_policy_without_executor_changes` registers a second policy and sees it applied, with the executor untouched.
4. **#689's behaviour is unchanged.** ✅ IDLE, changed-pair, and failed-read tests all survive with **no rewritten assertions**. The old "unchanged pair" test was not dropped — it was upgraded to `test_renamed_cycle_with_unchanged_pair_keeps_its_posting_path`, asserting the same posting path *through a rename*.

## Deliberately not done

`auth_preflight` and `quota_preflight` were **not** folded into the registry. The issue floats it as an option; they already apply to every Cycle, it is not free, and it would make an already-large diff unreviewable. Worth a follow-up ticket if the framework proves itself.

---

## ⚠️ Merge order — this PR must land AFTER #676

`0052_cycle_execution_policy_key` (this PR) and `0051_open_ask_terminal_states`
([#676](https://github.com/Illospace/illospace/pull/676)) both declare
`down_revision = "0050_scheduler_cold_start"`. They are parallel alembic heads.
Each one passes CI **alone**, because neither branch can see the other's file.
The second one to merge breaks `main`.

Measured on `main` (`031a692b`) with both migration files present:

```
$ alembic heads
0051_open_ask_terminal_states (head)
0052_cycle_execution_policy_key (head)

$ pytest tests/test_alembic_config.py::test_alembic_revision_chain_has_single_head
FAILED — Expected 1 head, got 2
```

**What to do:** merge [#676](https://github.com/Illospace/illospace/pull/676)
first. Then this PR's `down_revision` needs one line changed to
`"0051_open_ask_terminal_states"` (and the matching `Revises:` docstring line —
`test_alembic_revision_headers_match_identifiers` checks that the two agree).
R3 makes that edit and re-runs CI on its next run after #676 lands; the note is
here so the constraint is not lost if the PR is merged out of band.


redawear:
**Hold this PR — it needs two changes before it can merge.** Both found by an orchestrator merge-proof of all 17 open PRs against `origin/main@031a692b`, neither visible to CI.

1. **Alembic second head.** This PR's `0052_cycle_execution_policy_key` and #676's `0051_open_ask_terminal_states` both declare `down_revision = "0050_scheduler_cold_start"`. Merging both leaves two heads and the single-head test fails. The fix is a one-line re-point of `down_revision` to `0051_open_ask_terminal_states` — it cannot be made now, because this branch does not contain the `0051` file and its DB lane would go red. R3 does it once #676 merges.

2. **Collides with train B.** Train B (#678 → #686) rewrites `brain/systems/cycles/service.py` (**159+/202−**); this PR adds into the region it rewrote. Conflicts on `service.py` and `contracts.py`. R3 re-applies the additions on top of train B's refactored `_async_start_admitted_cycle_run` after #686 lands.

Merge order: train A (#676) and train B (#686) first, then R3 hands this back green. Proof: `state/evidence/r3-merge-proof-20260806T0100.md`.

redawear:
**Hold lifted — this is ready, as the top of a 2-PR train.**

The hold had two reasons. Both are resolved.

**1. Train B rewrote the region this PR adds into.** #678 merged, so the parent (#693) took `main` first, and this branch then merged the updated parent — a merge, not a rebase, so nothing was force-pushed. Four promotion-gate tests now assert `PromotionReadinessOutcome` (`UNCHANGED`/`EVALUATE`/`IDLE`/`UNAVAILABLE`) instead of the parallel `CycleExecutionEffect` shape #678 removed. Display-name ambiguity tests were replaced by durable-key validation plus unknown-key failure coverage — a renamed cycle is now valid, because the durable key is authoritative. That is this PR's whole point.

**2. Two Alembic heads.** `0052_cycle_execution_policy_key` declared `down_revision = "0050_scheduler_cold_start"`, and `0051_open_ask_terminal_states` landed in `main` claiming the same parent. Re-pointed onto 0051, docstring `Revises:` line included. Verified, not assumed:

```
$ alembic heads
0052_cycle_execution_policy_key (head)
```

Exactly one head. The repo's own single-head test passes.

**Verification (orchestrator-run, outside the worker sandbox):**
- Backend full CI selection: `4823 passed / 11 skipped / 0 failed`.
- Frontend lane — the one that broke last time on this same PR: `svelte-check` **6697 files, 0 errors, 0 warnings**; `npm test` 248 passed; build OK. The preview cycle factory in `cycles/+page.svelte` was the miss then; it carries the new policy key now.

**Merge order: #693 first, then this one.** GitHub merges the stack in one click.

redawear:
Superseded by #711 — one branch with all four, conflicts resolved (the two competing alembic `0052` revisions are now a single `0052 -> 0053` chain). Merge #711 and close this.

PR #708, opened 2026-08-06
fix(meetbot): report an empty meeting session while the humans are still there (#704)

redawear:
Closes #704

## What this fixes

On 2026-08-05 Illo joined a Google Meet, sat there **two hours**, captured zero
captions and observed zero participants, and said nothing until the meeting was
over. Reda asked *"are you? really?"* 37 seconds in and was right. The meeting is
unrecoverable — no transcript, no decisions, no action items.

**The detection already existed. The delivery did not.**

`SessionManager._warn_if_no_captions` already fired at 90 seconds and called
`record.add_warning(...)`. But that only mutated the in-memory `SessionRecord`,
and the record shipped exactly once — at the end — as a terminal
`meeting_transcript` envelope. So meetbot noticed in 90 seconds and stayed silent
for two hours.

This PR gives that existing signal a mid-session delivery path, and adds the
participant rule it never had.

## What changed

- **`meetbot/session_health.py` (new)** — a `SessionHealthMonitor` owning the task,
  clock, deadlines, wake event, sequence counter and warning latches, behind
  `start()` / `on_admitted()` / `on_caption_observed()` / `stop()`. `sleep` and
  `monotonic` are injectable, so timing is deterministically testable.
- **`meetbot/session.py`** — the one-shot caption-warning task is replaced by that
  monitor. `_ManagedSession` carries one `health_monitor` field.
- **`meetbot/callback.py`** — `MeetingWebhookSender` / `MeetingWebhookCallback`
  (renamed from `Completion*`, which no longer described what they carry) gain
  `send_health` beside `send_transcript`, reusing the existing bearer auth,
  idempotency key, bounded retry and dead-letter behavior.
- **`brain/systems/meetings/inbound.py`** — a new non-terminal
  `meeting_session_health` envelope kind, registered through the **existing**
  `register_inbound_envelope_handler` registry. Warnings route to the inviting
  thread through the same `_resolve_meeting_identity` / `_meeting_target`
  resolution the terminal report already uses.
- **`brain/systems/inbound/surface_admission.py`** — split into
  `prepare_surface_envelope` / `admit_prepared_surface_envelope` /
  `complete_prepared_surface_envelope`. A healthy observation completes through the
  neutral path; only a validated *warning* enters run admission. The shared path
  always admits with a concrete `run_id`.
- **`meetbot/config.py`** — `caption_warning_seconds` was declared on the dataclass
  but **`from_env()` never read it**, so the knob was silently unconfigurable.
[...]

PR #710, opened 2026-08-06
fix(ci): make the CI result gate fail closed, with needs as the job registry (#709)

### LINKED ISSUE #709: [ci] CI result reports success when the job-selection step never ran — a PR can go green with zero tests executed

## Problem

`Brain CI`'s aggregate gate, the `CI result` job, reports **success when no test ever ran**.

The gate is a deny-list. It walks the four `needs.*.result` values and fails only on the two words it knows:

```yaml
if [ "$result" = "failure" ] || [ "$result" = "cancelled" ]; then
  echo "A selected CI job failed or was cancelled."
  exit 1
fi
```

Anything else passes — including the state a dead `changes` job leaves behind. `changes` is the job that selects which suites run. When it does not complete, the three suites see `needs.changes.outputs.* == ''`, skip, and the gate reads three benign `skipped` values plus one value it does not recognise, and prints "Selected CI jobs passed."

This is the same class as #682: an absent result reads as "fine" instead of "never tested". #682 fixed the trigger so stacked PRs get checks at all; this is the gate itself failing open one level down.

## Evidence

PR #708, head `42a678d71060df712a07657aa14074f2b0404c43`, run [31118459372](https://github.com/Illospace/illospace/actions/runs/31118459372), 2026-08-06T16:03:43Z:

| job | conclusion |
|---|---|
| Detect changed areas | **cancelled** after 15m03s — annotation: *"The job was not acquired by Runner of type hosted even after multiple attempts"* |
| Backend fast suite | skipped |
| Frontend check and build | skipped |
| Backend DB suite | skipped |
| **CI result** | **success** |

Zero of ~4800 backend tests executed on that commit, and the PR's aggregate check is green. The failure was GitHub runner allocation, not code — which is exactly why it will recur, and why the gate must not depend on the infrastructure it is checking.

## Proposed fix

Make the gate an **allow-list** in `.github/workflows/brain-ci.yml`:

1. `needs.changes.result` must equal `success`. `changes` carries no `if:` condition, so it runs on every event — any other result (failure, cancelled, skipped, empty) means the selection never happened and nothing downstream can be trusted.
2. Each suite's result must be `success` or `skipped`, and `skipped` is only admissible once (1) holds. Fail on any unrecognised value rather than passing it.
3. Say which job produced which result in the failure message, so the next occurrence is diagnosable from the gate's own log instead of the API.

`Deployment Smoke` lost its runner in the same event ([31118459429](https://github.com/Illospace/illospace/actions/runs/31118459429)) and surfaced red on the PR, so it needs no change — this is specifically about the job whose purpose is to answer "did the tests pass".

## Acceptance checks

1. A run where `changes` is cancelled or fails ends with `CI result` **red**, not green. Simulate by forcing the job to fail, or by unit-testing the shell guard against the four result vectors.
2. An ordinary PR that touches only `frontend/**` still passes with the two backend suites `skipped` — the fix must not turn legitimate path-filter skips into failures.
3. A PR with a genuinely failing backend suite still fails, unchanged.
4. The gate's log names each job and its result on both the passing and the failing path.

## Related

- #682 / PR #683 — stacked PRs ran no CI at all; same failure shape, one level up.
- #708 — the PR that exposed this. Its head commit still has no CI coverage.

redawear:
Closes #709.

## What was wrong

`Brain CI`'s `CI result` job — the single check a human reads to decide "did the tests pass" — reported **success when no test ran**.

The guard was a deny-list. It failed only on the two literal strings `failure` and `cancelled`, so every other value passed. `Detect changed areas` is the job that selects which suites run; when it dies, the three suites see empty selection outputs, skip, and the gate reads three benign `skipped` values and prints "Selected CI jobs passed".

Live on PR #708, head `42a678d`, run [31118459372](https://github.com/Illospace/illospace/actions/runs/31118459372), 2026-08-06T16:03Z:

| job | conclusion |
|---|---|
| Detect changed areas | **cancelled** — *"The job was not acquired by Runner of type hosted even after multiple attempts"* |
| Backend fast suite | skipped |
| Frontend check and build | skipped |
| Backend DB suite | skipped |
| **CI result** | **success** |

Zero of ~4800 tests ran and the aggregate check went green. The trigger was GitHub runner allocation, so it recurs on its own schedule.

Same family as #682 (stacked PRs ran no CI at all), one level down: there the checks were absent, here they are present and lying.

## What this does

The guard becomes an allow-list, extracted to `scripts/ci_result_gate.sh` so it can be tested:

- `changes` must be exactly `success`. It carries no `if:` condition, so it runs on every triggering event — any other value means job selection never happened and no downstream `skipped` can be trusted.
- Every other job must be exactly `success` or `skipped`, and `skipped` only counts because the rule above already proved the selection ran and deliberately did not pick that area.
- Every job and its result is printed before the verdict, on both paths, so the next occurrence is diagnosable from this job's own log instead of the REST API.

**`needs:` is the job registry — there is no second list.** The whole `needs` context is handed to the script as `NEEDS_JSON: ${{ toJSON(needs) }}` and the job set is derived from it. Add a job to `needs:` and it is validated automatically; there is nowhere to forget to add it. That second fail-open (in the job *list*, rather than the job *results*) is what the first draft of this fix still had, and `test_new_job_is_checked_without_adding_its_name_to_the_gate` is the regression that pins it shut.

## See it yourself — no diff reading required

The gate is one shell command. On this branch:

```bash
NEEDS_JSON='{"changes":{"result":"cancelled"},"backend-fast":{"result":"skipped"},"frontend":{"result":"skipped"},"db-contract":{"result":"skipped"}}' bash scripts/ci_result_gate.sh ; echo "exit=$?"
```

That is the exact vector that shipped green this morning. It now prints each job, then `CI result failed: changes=cancelled; rule: the changes result must be exactly success.` and `exit=1`.

```bash
NEEDS_JSON='{"changes":{"result":"success"},"backend-fast":{"result":"success"},"frontend":{"result":"skipped"},"db-contract":{"result":"success"}}' bash scripts/ci_result_gate.sh ; echo "exit=$?"
```

An ordinary backend-only PR: `Selected CI jobs passed.` and `exit=0`.

## Acceptance checks

1. `changes` cancelled, failed, skipped, or absent → `CI result` is **red**, whatever the suites say.
2. A PR touching only `frontend/**` still passes with both backend suites `skipped` — legitimate path-filter skips must not become failures.
3. A genuinely failing backend suite still fails, unchanged.
4. The gate's log names each job and its result on the passing path and the failing path.

All four are asserted in `tests/test_ci_result_gate.py`, 14 cases: the four rules above, an unlisted fifth job that fails, an empty result string, a missing `changes` key, and unset / empty / malformed `NEEDS_JSON`.

## Verification

Every lane this diff selects, run locally on `ccea9da`:

| lane | result |
|---|---|
| Backend fast suite | **4845 passed**, 11 skipped (73s) |
| Backend DB suite | **80 passed**; `alembic upgrade head` clean against pgvector/pg16 |
| Frontend check and build | **248 passed**; `svelte-check` 6697 files, **0 errors 0 warnings**; build ok |
| Cycle context admission | ok |

The frontend lane is not incidental: `.github/workflows/brain-ci.yml` sits in both the `backend` and `frontend` paths-filters, so this PR selects all three lanes and exercises the new gate on a full fan-out.

## Assumptions

- `changes` genuinely always runs. Verified: the job has no `if:` — only its *steps* are conditioned on `github.event_name != 'workflow_dispatch'` — so `success` is the only correct result on every trigger.
- The added `actions/checkout@v4` step in `ci-result` does not weaken the gate. For `pull_request` events the workflow file itself already comes from the PR's merge ref, so the script is no more PR-authored than the inline shell it replaces. Independently reviewed and cleared.
- `jq` is preinstalled on GitHub-hosted `ubuntu-latest` runners, and on this machine. A contributor without `jq` sees `tests/test_ci_result_gate.py` fail rather than skip — worth a `pytest.importorskip`-style guard if that ever bites, but not worth pre-empting.

## Not addressed here

- **#708's own head commit still has no CI coverage.** This PR fixes the gate; it cannot retroactively test that commit. #708 needs a real CI run before it merges — [commented there](https://github.com/Illospace/illospace/pull/708#issuecomment-5207518953).
- `Deployment Smoke` lost its runner in the same event and correctly surfaced red on the PR, so it needs no change.

### DIFF
--- a/.github/workflows/brain-ci.yml
+++ b/.github/workflows/brain-ci.yml
@@ -211,17 +211,14 @@ jobs:
     if: always()
 
     steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+
       - name: Check selected CI jobs
         run: |
-          for result in \
+          set -euo pipefail
+          bash scripts/ci_result_gate.sh \
             "${{ needs.changes.result }}" \
             "${{ needs.backend-fast.result }}" \
             "${{ needs.frontend.result }}" \
             "${{ needs.db-contract.result }}"
-          do
-            if [ "$result" = "failure" ] || [ "$result" = "cancelled" ]; then
-              echo "A selected CI job failed or was cancelled."
-              exit 1
-            fi
-          done
-          echo "Selected CI jobs passed."
--- a/.github/workflows/brain-ci.yml
+++ b/.github/workflows/brain-ci.yml
@@ -215,10 +215,6 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Check selected CI jobs
-        run: |
-          set -euo pipefail
-          bash scripts/ci_result_gate.sh \
-            "${{ needs.changes.result }}" \
-            "${{ needs.backend-fast.result }}" \
-            "${{ needs.frontend.result }}" \
-            "${{ needs.db-contract.result }}"
+        env:
+          NEEDS_JSON: ${{ toJSON(needs) }}
+        run: bash scripts/ci_result_gate.sh
--- a/scripts/ci_result_gate.sh
+++ b/scripts/ci_result_gate.sh
@@ -2,35 +2,91 @@
 
 set -euo pipefail
 
-changes_result=${1-}
-backend_fast_result=${2-}
-frontend_result=${3-}
-db_contract_result=${4-}
+needs_json=${NEEDS_JSON-}
 
-printf 'changes=%s\n' "$changes_result"
-printf 'backend-fast=%s\n' "$backend_fast_result"
-printf 'frontend=%s\n' "$frontend_result"
-printf 'db-contract=%s\n' "$db_contract_result"
+if [ -z "$needs_json" ]; then
+  if [ "${NEEDS_JSON+x}" = "x" ]; then
+    input_state="<empty>"
+  else
+    input_state="<unset>"
+  fi
 
-if [ "$changes_result" != "success" ]; then
   printf '%s\n' \
-    "CI result failed: changes=$changes_result; rule: changes must be exactly success because job selection must complete."
+    "CI result failed: NEEDS_JSON=$input_state; rule: NEEDS_JSON must contain a non-empty valid JSON object."
   exit 1
 fi
 
-check_suite_result() {
-  local job_name=$1
-  local result=$2
+if ! jq -e -s 'length == 1 and (.[0] | type == "object")' \
+  >/dev/null 2>&1 <<<"$needs_json"; then
+  printf '%s\n' \
+    "CI result failed: NEEDS_JSON=<invalid>; rule: NEEDS_JSON must contain a non-empty valid JSON object."
+  exit 1
+fi
 
-  if [ "$result" != "success" ] && [ "$result" != "skipped" ]; then
-    printf '%s\n' \
-      "CI result failed: $job_name=$result; rule: suite result must be exactly success or skipped."
-    return 1
-  fi
-}
+jq -r '
+  def job_result:
+    if type == "object" then
+      if has("result") then .result else "" end
+    else
+      ""
+    end;
 
-check_suite_result "backend-fast" "$backend_fast_result"
-check_suite_result "frontend" "$frontend_result"
-check_suite_result "db-contract" "$db_contract_result"
+  to_entries
+  | sort_by(.key)[]
+  | (.value | job_result) as $result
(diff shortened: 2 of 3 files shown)


redawear:
Superseded by #711 — one branch with all four, conflicts resolved (the two competing alembic `0052` revisions are now a single `0052 -> 0053` chain). Merge #711 and close this.

PR #740, opened 2026-08-08
fix(worker): bound the drain when a health watchdog exits for restart

### LINKED ISSUE #739: [platform] Worker health self-exit cannot exit: infinite drain waits forever on a wedged run

The worker's health watchdogs log `exiting for restart` and raise `SystemExit(1)`. The process then hangs in its own shutdown and never exits, so `restart: unless-stopped` never fires. The box keeps a green `docker ps` with **zero claiming capacity**.

This happened live on illo-dev on 2026-08-08 and cost 45 minutes of total queue stoppage.

## Live evidence (2026-08-08, illo-dev)

| time (UTC) | fact |
|---|---|
| 17:05:19 | run **15747** starts (`recipe=worker`), last `updated_at` 17:18, still `running` at 20:29 |
| 19:45:16 | run 15792 queued |
| 19:46:18 | `[cortex.worker] ERROR agent-run worker queue stalled; exiting for restart` |
| 19:50:30 | last worker log line ever — the other in-flight runs finish |
| 20:29 | container `Up 4 hours`, `RestartCount=0`, `python -m brain.systems.cortex.worker` still alive, 18 runs queued, 1 stuck `running` |

Nothing after 19:46:18 logged a drain timeout, a supervisor-teardown warning, or `agent-run worker stopped`. The shutdown sequence started and stopped inside `stop_runner`. `illospace-scheduler-1` ticked green every 15s with `"reclaimed": 0` throughout, and `/api/health` stayed 200 — the outage was invisible to every health surface (more evidence for #737).

Recovered by hand: `docker rm -f illospace-worker-1` + `compose up -d --no-deps worker`. The fresh worker claimed the whole backlog in seconds and the reaper settled 15747.

## Mechanism

Three defects compose into the deadlock.

**1. A wedged run's in-flight id is never released.** `runner.py:1391` `_run_queued_once_async` adds the id to `_in_flight_run_ids` at claim time and removes it in a `finally` *inside the coroutine*. A task blocked in a sync tool call (run 15747 — a leftover headless Chromium was still running in the container) never reaches that `finally`, and cancellation cannot preempt a blocking executor call. `_supervisor_async_loop`'s teardown (`runner.py:1508`) clears `_runner_slots` but never touches `_in_flight_run_ids`, so `runner_in_flight_count()` stays ≥ 1 forever.

**2. The drain gate has no floor.** `stop_runner` (`runner.py:1590`) loops `while runner_in_flight_count() > 0` and, when `drain_timeout_seconds is None`, sleeps 0.2s forever. Production sets `ILLO_AGENT_RUNNER_DRAIN_TIMEOUT_SECONDS=infinity`, so `_shutdown_drain_timeout_seconds()` returns `None`.

**3. The self-restart path reuses the deploy drain policy.** `worker.py:265` runs one `finally` for every exit reason. An infinite drain is *correct* for a deploy SIGTERM — the handoff worker is already claiming, so long runs may finish at leisure. It is *exactly wrong* for a health self-exit: no handoff exists, nothing else claims, and the whole point of the exit is to restore claiming capacity. So `_terminate_process(exit_code)` at `worker.py:281` is never reached.

The invariant the code does not hold today: **a health self-exit must always terminate the process.** Three watchdogs depend on it — `runner supervisor unhealthy` (`worker.py:216`), `queue stalled` (`worker.py:230`), `cycle scheduler wedged` (`worker.py:252`).

## Acceptance criteria

- [ ] A health self-exit drains with a bounded timeout that ignores `ILLO_AGENT_RUNNER_DRAIN_TIMEOUT_SECONDS=infinity`; runs still in flight at the deadline are requeued through the existing `_recover_timed_out_runs` path.
- [ ] A deploy SIGTERM keeps the unbounded drain — the handoff contract does not change.
- [ ] A hang anywhere in the self-restart shutdown sequence (drain, requeue, cycle-scheduler stop, supervisor join) still ends in process exit, so the container restarts.
- [ ] Regression test: a permanently non-zero `runner_in_flight_count()` plus an infinite drain setting still reaches `_terminate_process`.

## Open question (not blocking the fix)

Why the queue stalled at 19:45 is unproven — run 15792 went unclaimed for 45s+ while ~5 of 10 slots were busy. The container logs were destroyed by the recovery, so the trigger cannot be recovered. Either it was a real claim stall or the stall detector fired spuriously; both are survivable once the self-restart can actually restart. Worth a follow-up if `queue stalled` shows up again.

Related: #556 (closed — the handoff-worker variant of "nothing can restart the worker"), #527 (closed — reboot variant), #737 (no view answers "is Illo healthy right now").

redawear:
Closes #739.

## What broke

On 2026-08-08 at 19:46:18 UTC the queue-stall watchdog on illo-dev logged `agent-run worker queue stalled; exiting for restart` — and the worker never exited. Container `Up`, `RestartCount=0`, the Python process alive, 18 runs queued, **zero claiming capacity for 45 minutes** while `docker ps`, `/api/health` and the scheduler's 15-second tick all stayed green.

The three health watchdogs exist to escape exactly this state. They raise `SystemExit(1)` and trust `restart: unless-stopped` to bring a fresh worker back. That contract holds only if the process actually dies.

It could not. `main()` runs one `finally` for every exit reason and always drains with `_shutdown_drain_timeout_seconds()`, which production sets to `infinity`. `stop_runner` then loops `while runner_in_flight_count() > 0` with no floor. Run 15747 had wedged at 17:18 inside a blocking tool call (a leftover headless Chromium was still running in the container), so its id never left `_in_flight_run_ids` — that decrement lives in a `finally` inside the coroutine, and cancellation cannot preempt a blocked executor call. The drain waited forever and `_terminate_process` was never reached.

## The fix

**A health self-exit must always terminate the process.** Two changes make that true:

1. **The self-restart path gets its own bounded drain** (`ILLO_AGENT_RUNNER_SELF_RESTART_DRAIN_TIMEOUT_SECONDS`, default 60s, never resolves to "forever" — `float("infinity")` is rejected explicitly, since copying the deploy value across would restore the deadlock). Runs still in flight at the deadline flow through the existing `_recover_timed_out_runs` requeue path, so the fresh worker picks them up.
2. **A watchdog timer covers the rest of the shutdown.** Draining, requeuing, stopping the cycle scheduler and joining the supervisor can each block on the same wedged run or database the worker is escaping. The timer force-exits after the drain budget plus 30s.

**The deploy path is untouched.** A SIGTERM still drains indefinitely, because there the handoff worker is already claiming and long runs may finish at leisure. The three watchdogs now share `_exit_for_restart`, which records *why* the process is leaving so the shutdown can tell "a peer is covering the queue" from "nobody is".

## What this deliberately does not change

The `_in_flight_run_ids` leak stays. That set is a truthful record of runs this process claimed and has not finished, and the deploy swap gate reads it as truth — clearing it on teardown would make a drain report "done" while a run is still executing in a wedged thread. Bounding the *wait* is the right owner of the invariant; fixing the bookkeeping is not.

## Verification

- 6 new tests in `tests/test_cortex_worker.py`, including the live shape: a `stop_runner` that refuses an unbounded timeout, driven through the queue-stall watchdog with `ILLO_AGENT_RUNNER_DRAIN_TIMEOUT_SECONDS=infinity` set. It fails against the pre-fix code.
- `test_deploy_sigterm_keeps_the_unbounded_drain` pins the handoff contract.
- `tests/test_cortex_worker.py tests/test_worker_swap_deploy.py tests/test_stale_run_reaper.py tests/test_agent_run_state_machine.py` — 106 passed, 8 skipped.
- The box was recovered by hand at 20:31 UTC (`docker rm -f` + `compose up -d --no-deps worker`); the fresh worker claimed the whole backlog in seconds and the reaper settled run 15747. Queue is empty as of 20:40.


redawear:
## Codex adversarial review (xhigh) — 6 findings, all addressed

Cross-family review caught that the first pass had the guarantee **keyed the wrong way round**.

| # | severity | finding | fix |
|---|---|---|---|
| 3 | HIGH | `_restarting_for_health` was set by three explicit branches only, so any other fatal path — including an exception *inside* those health checks — still took the deploy's infinite drain | Inverted: only the signal handler earns an unbounded drain, because only a deploy verifies a claiming handoff before signalling. Everything else is bounded |
| 2 | HIGH | `1e308` passes `isfinite`, defeats the drain, and makes `threading.Timer` raise `OverflowError` | Added a 3600s ceiling |
| 1 | HIGH | The watchdog was armed *after* a log line, cancelled before the last three shutdown steps, and its callback logged through the logging lock — a blocked log pipe defeated all three | Armed before the first log, never cancelled (daemon timer dies with the process), and the killer writes to fd 2 instead of taking the logging lock |
| 5 | MEDIUM | Compose env is a whitelist, so the new tunable was dead on arrival | Wired `ILLO_AGENT_RUNNER_SELF_RESTART_DRAIN_TIMEOUT_SECONDS` into `docker-compose.yml` |
| 6 | MEDIUM | The hang test passed with no watchdog at all — the fake hang returned on its own after 5s | It now records *what released the block* and fails unless the timer did it. Verified: neutering `_arm_shutdown_watchdog` makes it fail |
| 4 | MEDIUM (speculative) | A 60s drain can requeue healthy long runs on an unrelated cycle-scheduler exit | Kept 60s and raised the shutdown grace 30s → 60s. Requeue is fenced by execution token, and the stale reaper already requeues at 300s, so this is existing accepted behaviour — total queue stoppage is the worse outcome. Now tunable per finding 5 |

Codex also confirmed two things I had reasoned about and left alone: the deploy SIGTERM contract is unchanged (it verifies a claiming handoff at `compose-runtime-lib.sh:566` before signalling), and clearing `_in_flight_run_ids` in supervisor teardown would be wrong — it would discard the run ids needed for recovery and let the deploy drain finish while work is still executing.

Both new regression tests are proven against the designs they replace:
- `test_an_unclassified_crash_also_drains_with_a_floor` fails on 2c200053 (the positive-flag design)
- `test_health_exit_terminates_even_when_the_shutdown_sequence_hangs` fails when the watchdog is neutered

123 passed, 8 skipped locally.

PR #753, opened 2026-08-10
fix(knowledge): break circular import on immediate indexing (#751)

### LINKED ISSUE #751: [knowledge] Circular import kills immediate memory indexing on every commit — 85 silent failures a day, masked by the 30-minute sync

## Problem

Every memory node commit fails its immediate knowledge indexing, because `brain/systems/knowledge/service.py` cannot be imported without tripping a circular import.

Found incidentally while investigating disk usage on illo-dev (#750). The worker log carries 85 occurrences of this error over the container's 23-hour lifetime — one per committed memory node, continuously:

```
2026-08-09 21:52:52 [brain.systems.reconstructive_memory.ingestion] ERROR
  Immediate knowledge indexing failed for committed memory node
```

## The cycle

```
brain/systems/knowledge/service.py:30       from ...connectors.memory import MemoryConnector
  -> brain/systems/knowledge/connectors/__init__.py:4   from ...connectors.github import GitHubConnector
    -> brain/systems/knowledge/connectors/github.py:41  from brain.systems.knowledge.service import RAW_TEXT_MAX_CHARS
```

`RAW_TEXT_MAX_CHARS` is defined at `brain/systems/knowledge/service.py:47` — **after** the connector import on line 30. So when `service` is imported first, line 30 pulls in the connectors package, `github.py` asks the half-built `service` module for a name that does not exist yet, and the import fails:

```
ImportError: cannot import name 'RAW_TEXT_MAX_CHARS' from partially initialized
module 'brain.systems.knowledge.service' (most likely due to a circular import)
```

## Severity: degraded, not data loss

The scheduled `knowledge_index_sync` job (every 30 minutes) is healthy — 48 of 48 runs `settled_success` in the last 24 hours — and it picks up the nodes that immediate indexing dropped. So the real cost is:

- Up to 30 minutes of staleness between committing a memory and its being searchable.
- 85 ERROR-level stack traces per day in the worker log, which train readers to ignore errors.
- The immediate-indexing path is dead code in practice; nobody has noticed because the fallback hides it.

This is a different failure from #627 (which was `knowledge_index_sync` itself failing and freezing the watermark). That job is fine now; this is the immediate path.

## Fix

Move `RAW_TEXT_MAX_CHARS = 20_000` above the connector imports in `service.py`, or move the constant into a leaf module that both `service.py` and `connectors/github.py` can import without a cycle. The second option is more durable — the cycle will come back the next time a connector needs a constant from `service`.

## Acceptance criteria

- Importing `brain.systems.knowledge.service` first, in a clean interpreter, raises no ImportError. A test asserts this so the cycle cannot silently return.
- `Immediate knowledge indexing failed for committed memory node` no longer appears in the worker log.
- A newly committed memory node is searchable without waiting for the next `knowledge_index_sync` tick.

redawear:
Closes #751

## What was wrong

`brain/systems/knowledge/service.py` defined `RAW_TEXT_MAX_CHARS`, and
`brain/systems/knowledge/connectors/github.py` imported that name back from
`service`. Because `service` imports the connector package first, importing
`brain.systems.knowledge.service` in a clean interpreter raised:

```
ImportError: cannot import name 'RAW_TEXT_MAX_CHARS' from partially
initialized module 'brain.systems.knowledge.service'
(most likely due to a circular import)
```

Under pytest the cycle is invisible — collection has already imported the
package by the time any test runs — so it only failed in the worker, which is
why it showed up as ~85 silent `Immediate knowledge indexing failed for
committed memory node` lines a day, masked by the 30-minute sync.

## The fix

Move the constant one level down to `brain/kernel/config.py`, which already owns
the knowledge indexing tunables (`KNOWLEDGE_EMBEDDING_DIM`,
`KNOWLEDGE_CONNECTOR_BATCH_SIZE`, `KNOWLEDGE_GITHUB_REPOSITORIES`). Both modules
now import it from there, so no edge points back up.

Three source lines changed. `service.py` still re-exports the name in its
`__all__`, so `from brain.systems.knowledge.service import RAW_TEXT_MAX_CHARS`
keeps working for existing callers (`tests/test_knowledge_index.py` relies on
this).

## How to check it

The repro is one command, and it fails on `main`:

```
python3 -c "import brain.systems.knowledge.service"
```

The regression test runs it the only way that can catch it — in a **subprocess**
with a clean interpreter, for both entry points. An in-process `import` proves
nothing here.

```
venv/bin/python -m pytest tests/test_knowledge_imports.py -q
```

## Acceptance

- [x] `import brain.systems.knowledge.service` first, in a clean interpreter, raises no ImportError
- [x] A test asserts it so the cycle cannot silently return (`tests/test_knowledge_imports.py`, subprocess, both entry points)
- [x] No other importer of `RAW_TEXT_MAX_CHARS` breaks — verified by grep, and the name stays exported from `service`
- [ ] `Immediate knowledge indexing failed…` disappears from the worker log — only observable after deploy

## Assumptions

- `brain/kernel/config.py` is the right home because it already holds the other
  knowledge tunables. The alternative — a new constants module — would add a
  leaf the repo does not otherwise have.
- The last acceptance line and "a newly committed node is searchable without
  waiting for the next sync tick" are runtime observations. They cannot be
  proven from this branch; they follow from the import succeeding.


redawear:
Independent test evidence, run serially on the host with no other worker active (load ~4.4). The three concurrent worker runs earlier today all reported failures; those were contention, not code.

| run | result |
|---|---|
| clean `origin/main` @ `8abb5343`, throwaway worktree | **4805 passed**, 11 skipped, 0 failed, 0 errors |
| this branch @ `8760025d` | **4807 passed**, 11 skipped, 0 failed, 0 errors |

Command: `pytest tests/ -m "not requires_db and not requires_browser and not live_provider"`, 89s.

+2 is exactly the two new subprocess import tests. No pre-existing failure and no new one. The diff touches `brain/**` and `tests/**`, so the backend lane is the one that applies; the Postgres lane was not run locally, and this change adds no model and no migration.

Worth knowing for whoever merges: the earlier worker runs reported "1 failed, 5 errors" and "3 failed, 5 errors" on the same code. Those sets were disjoint and the clean baseline is fully green, so they were load and sandbox socket binds.

Issue #749, opened 2026-08-08
[deploy][worker] Confirmed-idle worker can hold upgrade for the 24-hour drain deadline


redawear:
## What happened

During the 2026-08-08 production reset to `8abb53432f56a14bca5e96a62a1e67c50f3a4b0c`, the canonical worker replacement path established that the old worker had zero active AgentRuns and sent it `SIGTERM` at `2026-08-08T23:06:25Z`.

The worker published its draining state and stopped claiming, but the process did not exit. `wait_for_worker_exit` printed its zero-active hint after 63 seconds and then continued waiting. It would not force the safe idle replacement until `COMPOSE_RUNTIME_WORKER_DRAIN_TIMEOUT_SECONDS`, whose production default is 86400 seconds.

After more than the normal shutdown window, the operator sent `SIGKILL` only to the confirmed-idle `illospace-worker-1`. The deploy immediately recreated the worker, and doctor passed with zero warnings. The new worker heartbeat was current and the live runtime snapshot was `good`.

## Why this happens

`wait_for_worker_exit` already checks the canonical worker snapshot every 30 seconds. Two consecutive zero-active checks only produce a hint; they do not bound the idle wait. Separately, a deploy-triggered signal lets `worker.py` use `ILLO_AGENT_RUNNER_DRAIN_TIMEOUT_SECONDS=infinity`, so a wedged shutdown has no internal watchdog.

This makes the safe no-active-run path capable of holding an operator deploy for up to 24 hours even though the worker has stopped claiming and there is no run to protect. This is a focused follow-up to the deploy-hang context in #527.

## Acceptance

- When the outgoing worker is draining and the canonical snapshot confirms zero active AgentRuns on consecutive checks, bound the remaining idle exit wait to a short configurable limit (for example 120 seconds), then force the existing safe replacement path.
- Do not shorten the drain while any active AgentRun is present.
- Keep restart-policy restoration crash-safe and finish with exactly one Compose-managed worker.
- Add a regression where the worker ignores `SIGTERM`, active count stays zero, and replacement happens at the idle deadline instead of the 86400-second deploy deadline.
- The deploy output must clearly state that no active runs are affected.

PR #788, opened 2026-08-10
refactor(deploy): split wait_for_worker_exit into covered and idle waits

### LINKED ISSUE #755: [deploy] Split wait_for_worker_exit — one function now carries two drain policies and three exit statuses

## Problem

`wait_for_worker_exit` in `deploy/scripts/compose-runtime-lib.sh` now serves two
different policies inside one function, selected implicitly by whether
`handoff_id` is empty:

- **empty** → confirmed-idle mode: arms a short idle deadline and can return **3**
- **non-empty** → covered mode: monitors the handoff worker and can return **2**

Three variables (`consecutive_zero_checks`, `idle_deadline`, `hint_printed`)
carry the idle state, and `hint_printed` gates deadline *arming*, not just
output. A reader has to simulate the loop to learn which states are reachable.

Both call sites use `|| drain_status=$?` with fallback branches, so the handoff
path would silently treat a status 3 as a drain timeout and the idle path would
treat a status 2 the same way. Nothing breaks today only because those
combinations are unreachable — an invariant held by construction, not stated
anywhere.

The active-run transition is also written twice (the same message and the same
three resets), so a future state change must edit both branches identically.

## Why it was deferred

This is deploy-critical shell. The repo has no CI minutes, and the tests drive it
through a simulator harness rather than a real Docker daemon, so a restructure
here cannot be proven safe the way an ordinary refactor can. On the #749 branch a
single-transition-block rewrite was attempted and **reverted**: in Bash it costs a
two-iteration pseudo-loop with the `sleep` buried in one branch, which reads worse
than the four duplicated lines it removes.

## Acceptance

- `wait_for_worker_exit` is split into a covered-worker wait and a no-handoff
  idle wait. The idle state variables live only in the idle function.
- Each function has one exit-status contract, and each call site handles it with
  an exhaustive `case` — no fallback branch that would mis-handle the other
  mode's status.
- The active-run transition is written once per function.
- Named constants replace the bare `1` / `2` / `3` if numeric statuses survive.
- `tests/test_worker_swap_deploy.py` still passes unchanged, including
  `test_idle_deadline_is_abandoned_and_rearmed_when_active_runs_reappear` and
  `test_active_run_at_idle_deadline_requires_two_fresh_zero_snapshots`.

Raised by a code-quality review of the #749 branch (`illo-qa/749-idle-drain-deadline`).

redawear:
Closes #755

## What changed

`wait_for_worker_exit` in `deploy/scripts/compose-runtime-lib.sh` carried two
drain policies in one function, chosen implicitly by whether `handoff_id` was
empty. Both call sites used `|| drain_status=$?` with a fallback branch, so the
handoff path would have treated a status `3` as a drain timeout and the idle path
would have treated a status `2` the same way. Nothing broke only because those
combinations were unreachable — an invariant held by construction and written
down nowhere.

It is now two functions with one contract each:

| function | 0 | 1 | 2 | 3 |
|---|---|---|---|---|
| `wait_for_worker_exit` (covered) | worker exited | drain timeout | cover lost | — |
| `wait_for_idle_worker_exit` (no handoff) | worker exited | drain timeout | — | confirmed-idle timeout |

- The idle state variables live only in the idle function.
- Named constants replace the bare `1` / `2` / `3`.
- Each call site uses an exhaustive `case`; an unexpected status stops the swap
  instead of guessing.
- The active-run transition is written once per function.

**Behavior is unchanged.**

## What I deliberately did NOT do

The ticket records that a *single shared transition block* was written and then
reverted on the `#749` branch, because in Bash it costs a two-iteration
pseudo-loop with the `sleep` buried in one branch. This PR does not reintroduce
that shape. The duplication between the two functions is intentional.

## Verification

- Repo CI command, run serially on an idle machine (load 3.4 before and after):
  → **4867 passed, 11 skipped**.
- `bash -n deploy/scripts/compose-runtime-lib.sh` passes. `shellcheck` is not
  installed on this machine, so it was not run.
- Cross-family code-quality review (Codex, thermo-nuclear rubric): **zero
  findings**, blocking or otherwise.

`tests/test_worker_swap_deploy.py` changed in one mechanical way only: the two
named idle tests now invoke `wait_for_idle_worker_exit` instead of
`wait_for_worker_exit worker-1 ''`. Their tick schedules and assertions are
untouched. That change is unavoidable when the function is split.

This is deploy-path shell, so the real exercise is a deploy:

1. Run a normal `./illo deploy` on illo-dev while **no** AgentRuns are active.
   The log should show `Worker: no interactive AgentRuns; signaling a graceful
   replacement.` and the swap should complete as before.
2. Run one with an active AgentRun so the handoff path is taken. The log should
   show the drain message and the handoff cover, unchanged from today.
3. Neither path should ever print `returned unexpected status` — that line is the
   new exhaustive-case guard and should be unreachable.

### DIFF
--- a/deploy/scripts/compose-runtime-lib.sh
+++ b/deploy/scripts/compose-runtime-lib.sh
@@ -483,8 +483,17 @@ record_worker_drain_timeout() {
     > "$COMPOSE_RUNTIME_WORKER_DRAIN_TIMEOUT_FILE" 2>/dev/null || true
 }
 
-# Waits for the outgoing worker to drain, and -- when a handoff worker is
-# covering for it -- for that cover to still exist.
+# Exit statuses for the covered-worker wait.
+COVERED_WORKER_EXITED_STATUS=0
+COVERED_WORKER_DRAIN_TIMEOUT_STATUS=1
+COVERED_WORKER_COVER_LOST_STATUS=2
+
+# Exit statuses for the no-handoff idle wait.
+IDLE_WORKER_EXITED_STATUS=0
+IDLE_WORKER_DRAIN_TIMEOUT_STATUS=1
+IDLE_WORKER_EXIT_TIMEOUT_STATUS=3
+
+# Waits for the outgoing worker to drain while a handoff worker covers it.
 #
 # Watching only the clock is what turned the 2026-07-28 illo-dev deploy into a
 # 2h16m outage. The handoff came up, passed the one-shot liveness check, claimed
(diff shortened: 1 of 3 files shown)


redawear:

redawear:

PR #789, opened 2026-08-10
refactor(failure-guard): narrow trigger registrations by mode

### LINKED ISSUE #773: [scheduler][refactor] Trigger mode tag does not narrow the trigger type — dispatch still needs casts

Split out of the cross-family code-quality review of the #769 branch
(`illo-qa/769-explicit-trigger-mode`). #769 made the mode explicit and closed the
"a method-set change silently rewires persistence" hole. This is the part of that
review deliberately left out of #769, because it is a type-level refactor rather
than a fix to the behavior #769 was about.

## Problem

`FailureGuardTriggerRegistration` carries `mode` and `trigger` as **independent**
fields, so the tag does not narrow the implementation type.

Consequences on `origin/main` after #769 lands:

- `brain/systems/failure_guard/core.py` needs three `cast(...)` calls in the
  dispatch path to assert what the mode already declared.
- `brain/app/scheduler/scheduler_failure_guard.py` — the registration alias
  discards `SchedulerFailureGuardTriggerImplementation`, which becomes unused.
  The casts there assert `evaluate_many` and `should_reset` without proof.

## Why it matters

A new scheduler trigger can pass registration validation with `evaluate` alone —
no `evaluate_many`, no `should_reset` — and then fail at dispatch, far from the
provider that registered it. #769's validator checks the *core* protocol methods;
it cannot see the scheduler's extra contract.

Adding a third mode later would also mean auditing several non-exhaustive
branches and every cast, rather than letting the type checker find them.

## Fix direction

Typed stateless and stateful registration variants with literal mode tags, then
define their union as the registration type. Consumer aliases bind each variant
to the consumer-specific protocols (`SchedulerFailureGuardTrigger` /
`SchedulerFailureGuardStatefulTrigger`, and the Cycle equivalents). This removes
the casts and makes the consumer's extra contract checkable at registration.

## Acceptance

- No `cast(...)` remains in the failure-guard dispatch path.
- A scheduler trigger registered without `evaluate_many` or `should_reset` fails
  at registration, with a test.
- `SchedulerFailureGuardTriggerImplementation` is either used by the registration
  types or deleted — not left unused.
- Both consumers keep working; behavior unchanged.

## Sequencing

Depends on #769. Base any branch for this on `illo-qa/769-explicit-trigger-mode`
until that merges.

redawear:
Closes #773

## What changed

`FailureGuardTriggerRegistration` carried `mode` and `trigger` as independent
fields, so the tag did not narrow the implementation type and dispatch needed six
`cast(...)` calls to assert what the mode had already declared.

- Adds `FailureGuardStatelessTriggerRegistration` and
  `FailureGuardStatefulTriggerRegistration` with literal `mode` tags.
  `mode` is `init=False`, so a caller can no longer construct a registration whose
  tag disagrees with its variant.
- `FailureGuardTriggerRegistration` is now their discriminated union.
- Consumer aliases bind each variant to the scheduler and Cycle protocols. The
  unused `SchedulerFailureGuardTriggerImplementation` alias is deleted.
- **All six casts are gone**; `assert_never` replaces them.
- A scheduler trigger missing `evaluate_many` or `should_reset` now fails at
  **registration**, not at dispatch far from the provider that registered it.

**Behavior is unchanged.** This is a type-level refactor.

## The review caught something and it is folded in

The cross-family code-quality review (Codex, thermo-nuclear rubric) found one
**blocking** issue in the first commit: the registry validated the scheduler
contract against a hand-written list of method names, so the protocols and the
check became two separate sources of truth. A maintainer adding a method to
`SchedulerFailureGuardTrigger` would leave the registration check silently
permissive — recreating the exact late failure this ticket exists to prevent.

Second commit fixes it: the scheduler protocols are `@runtime_checkable` and the
registry validates with `isinstance` against the protocol the mode selects. The
error message still names the missing methods, derived from `__protocol_attrs__`
rather than written out. The Cycle consumer carried no hand-listed methods and is
unchanged.

Runtime protocols check member presence, not signatures — the same guarantee the
previous `callable(getattr(...))` list gave. This is not a weakening.

## Verification

- Repo CI command, run serially on an idle machine (load 3.3 before, 5.3 after):
  → **4868 passed, 11 skipped** (one more than `main`: the new registration test).
- New test: `test_scheduler_registration_rejects_trigger_without_scheduler_contract`
  proves an evaluate-only scheduler trigger fails registration.
- The repo configures no mypy or pyright, so there is no static type gate to run.
  That is the point of the ticket: the types are now checkable if one is ever added.

There is no UI. The scheduler either starts or it does not:

1. Deploy to illo-dev and confirm `illospace-scheduler-1` comes up healthy —
   registry construction runs at import, so a mistake here fails fast and loudly.
2. Let one scheduled Cycle run and fail, then confirm the failure guard still
   latches and still resets exactly as it does today.

### DIFF
--- a/brain/app/scheduler/scheduler_failure_guard.py
+++ b/brain/app/scheduler/scheduler_failure_guard.py
@@ -11,7 +11,6 @@ from typing import (
     Protocol,
     Sequence,
     TypeAlias,
-    cast,
 )
 
 from sqlalchemy import and_, delete, func, or_, select
@@ -31,10 +30,11 @@ from brain.systems.failure_guard.core import (
     FailureGuardLifecycleEvent,
     FailureGuardLatch,
     FailureGuardStatefulTrigger,
+    FailureGuardStatefulTriggerRegistration,
+    FailureGuardStatelessTriggerRegistration,
     FailureGuardTrigger,
     FailureGuardTriggerKind,
     FailureGuardTriggerMode,
-    FailureGuardTriggerRegistration,
     require_failure_guard_registrations,
     FailureGuardTriggerResult,
     FailureGuardTriggerState,
@@ -512,11 +512,11 @@ class SchedulerFailureGuardStatefulTrigger(
     """One state-owning scheduler failure trigger."""
 
 
-SchedulerFailureGuardTriggerImplementation: TypeAlias = (
-    SchedulerFailureGuardTrigger | SchedulerFailureGuardStatefulTrigger
-)
 SchedulerRegisteredFailureGuardTrigger: TypeAlias = (
-    FailureGuardTriggerRegistration[SchedulerFailureGuardLifecycleContext]
+    FailureGuardStatelessTriggerRegistration[SchedulerFailureGuardTrigger]
+    | FailureGuardStatefulTriggerRegistration[
+        SchedulerFailureGuardStatefulTrigger
+    ]
 )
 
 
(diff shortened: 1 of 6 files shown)


redawear:

redawear:

Issue #790, opened 2026-08-10
[behavior-editor][refactor] Cycle policy snapshots are an untyped dict contract that five slices are about to build on


redawear:
> *This was generated by AI during triage.*

Raised by the cross-family code-quality review of the #782 branch
(`illo-qa/782-behavior-policy-command-module`), and deliberately left out of that
PR because it is a contract redesign rather than a fix to what #782 shipped.

## Problem

The behavior-policy boundary is typed at the edges and untyped in the middle.

- Every field on `CyclePolicyPatch`
  (`brain/systems/cycles/behavior_policy.py:77`) is `Any`.
- `EffectiveCyclePolicy.snapshot` (`:106`), previews, audit `before_snapshot` /
  `after_snapshot`, the preview digest, apply and revert all pass
  string-keyed dictionaries.
- The real schema exists only as a runtime field set inside `_validated_snapshot`
  (`:567`).

## Why it matters now rather than later

**Five slices are about to be built on this contract.** #783–#787 all read,
render, diff and edit these snapshots. Every one of them will hard-code the same
string keys, and a maintainer adding a policy field will have to find and update
several independent mappings by hand. No static check can find a missed one.

This is the "typed adapters later, without becoming a generic raw-row editor"
constraint from #738 story 30. #782 does not expose arbitrary Cycle columns — but
the raw dictionary boundary is the path by which it could.

## Not the forward-compatibility bug

The related read-back failure — an exact-field check that rejects stored
snapshots written before a field was added — is fixed inside #782 with a tolerant
decode plus an explicit snapshot version. **That fix is not this ticket.** This
ticket is the typed representation the tolerant decode should eventually be
implemented against.

## Fix direction

- A typed, versioned `CyclePolicySnapshot` with explicit JSON encode and
  compatibility decode. Dictionaries survive only at the JSON database boundary.
- Real types on `CyclePolicyPatch` fields.
- One omission sentinel owned by the patch contract.

## Acceptance

- Adding a Cycle policy field requires editing the snapshot type and nothing else
  that a type checker cannot find.
- Encode and decode are the only places that know JSON key names.
- Stored snapshots from every prior version still decode, with a test per version.
- No behavior change.

## Sequencing

Decide this **before #783** starts, so the API slice is not written against the
dictionary contract and then rewritten.

redawear:
> *This was generated by AI during triage.*

**Reda answered YES on 2026-08-10: this lands before #783.**

[...]

Issue #791, opened 2026-08-10
[behavior-editor] The behavior-change audit envelope is named generic but is structurally Cycle-only


redawear:
> *This was generated by AI during triage.*

Raised by the cross-family code-quality review of the #782 branch
(`illo-qa/782-behavior-policy-command-module`). Left out of that PR because the
answer is a design decision, not a mechanical fix.

## Problem

`BehaviorChangeAudit` (`brain/platform/db/models/cycle.py`, table
`behavior_change_audits`, migration `0059`) presents itself as a **generic**
policy audit envelope — it carries `policy_kind`, `target_type` and `target_id`
— but it is structurally Cycle-only:

- `cycle_revision_id` is **NOT NULL** with a foreign key to `cycle_revisions`.
- The model lives in the Cycle model module.

A future non-Cycle policy adapter therefore cannot write to this envelope without
either inventing a false Cycle revision or changing the schema.

## Why it is a decision, not a bug

#782's ticket explicitly wanted the two linked: *"CycleRevision stays the
execution revision; the behavior-change record is the human-readable audit
envelope. They are linked, so a past CycleRun can show both the execution
snapshot and the change that produced it."* The current shape honours that. The
generic column names are what create the mismatch.

Both resolutions are defensible and they point opposite ways:

1. **Make it honest and Cycle-specific.** Rename the table and the model, drop
   the generic `policy_kind` / `target_type` columns. A second policy kind gets
   its own envelope later. Smallest change, no speculative generality.
2. **Make it genuinely adapter-neutral.** Make `cycle_revision_id` nullable, or
   move the execution-revision link into adapter-owned metadata. Costs a
   migration now, avoids one later.

Do not pick by taste alone — pick by whether any policy kind other than Cycle is
actually planned. If none is, option 1 is correct and option 2 is speculative
generality.

## Acceptance

- The table's name and columns match what it can actually store.
- If it stays Cycle-linked, nothing in the schema advertises otherwise.
- If it becomes neutral, a second policy kind can write a row without a Cycle
  revision, proven by a test.
- Existing rows migrate either way.

## Sequencing

Not urgent for #783–#787, which are all Cycle policy. It becomes blocking the
first time a non-Cycle policy kind is added, and the migration is cheaper before
the table has production history.

Issue #787, opened 2026-08-10
[behavior-editor] Slice 6 — keyboard operability and small-screen rollback


redawear:
> *This was generated by AI during triage.*

Slice **6 of 6** of [#738](https://github.com/Illospace/illospace/issues/738),
sliced on Reda's instruction (2026-08-10). Read the PRD first.

## Scope

The editor is fully operable without a pointer, and usable in the situation it is
most needed: someone away from a desktop who has to roll a bad policy back.

- Review, apply, cancel, inspect history and revert are all reachable by
  keyboard.
- Focus order, labels, error association, and dialog focus return are correct.
- On a small screen the effective policy, the diff and the apply controls stay
  readable **without horizontal scrolling**.

## Why this is its own slice, not a polish pass

Urgent rollback is a mobile action. If the diff needs a sideways scroll on a
phone, the person rolling back cannot see what they are about to change — which
is exactly the moment the diff matters most.

## Acceptance

- Accessibility checks cover focus order, labels, error association, dialog focus
  return, and small-screen layout.
- Every action in slices 4 and 5 is keyboard-reachable.
- No horizontal scrolling on a phone-width viewport for policy, diff, or apply
  controls.

## Sequencing

Depends on slices 4 and 5.

Issue #786, opened 2026-08-10
[behavior-editor] Slice 5 — policy history and per-run snapshot inspection


redawear:
> *This was generated by AI during triage.*

Slice **5 of 6** of [#738](https://github.com/Illospace/illospace/issues/738),
sliced on Reda's instruction (2026-08-10). Read the PRD first.

## Scope

The reviewer's surface: how behavior got to where it is.

- History ordered by **effective version**, so behavior can be traced over time.
- Each entry shows actor type, actor identity, source, rationale, and affected
  fields — so a human change and an agent change are distinguishable at a glance.
- When a change came from a run, the entry identifies the originating run, so a
  user who asked Illo to change a Cycle can trace the result back to the request.
- Inspect the **exact policy snapshot a past CycleRun used**, so a result can be
  explained without guessing from current policy.

## Why the snapshot view matters

This is the payoff of the whole audit design. Today, explaining why a CycleRun
behaved a certain way means reading today's policy and hoping it did not change.
The run's own snapshot is the only honest answer, and slice 1 already stores it.

## Acceptance

- History is ordered by effective version and paginates.
- An agent-authored change is visibly distinct from a human one, with its
  originating run named when available.
- A past CycleRun shows both its execution snapshot and the change that produced
  it.
- Retired guidance stays readable in history.

## Sequencing

Depends on slice 4.

PR #792, opened 2026-08-10
feat(cycles): behavior-policy command module and immutable change audit (behavior editor slice 1)

### LINKED ISSUE #782: [behavior-editor] Slice 1 — behavior-policy command module and the immutable change audit envelope

Slice **1 of 6** of [#738](https://github.com/Illospace/illospace/issues/738),
sliced on Reda's instruction (2026-08-10). Read the PRD first — it holds the
problem statement, the 30 user stories and the implementation decisions. The
slices are cut by **contract seam**, not by user story, because the PRD already
names the seam.

This is the keystone. Nothing else in the set can start without it.

## Scope

One deep behavior-policy command module with a small interface:

- read effective Cycle policy
- preview a proposed change
- apply a reviewed change
- list history
- preview and apply a revert

Plus one immutable behavior-change audit envelope per apply, storing: workspace,
policy kind, target identity, monotonically increasing version, actor type and
identity, source reference, rationale, before and after snapshots, changed
fields, apply time, and an optional reverted-from reference.

## The rule the module exists to enforce

**API routes and agent tools call this module. Nothing mutates policy tables
directly.** An agent-authored change uses the same command service and the same
audit contract as a human edit, so Illo cannot bypass history or validation.

## Behavior it must guarantee

- **Apply is atomic**: update Cycle fields, retire guidance removed from the
  active set, add new immutable guidance rows, record a `CycleRevision`, record
  the audit change, and publish the existing Cycle change event — one transaction
  boundary.
- **Guidance text is immutable.** A wording edit retires the old row and creates
  a new one. Retirement changes effective state; it never deletes content.
- **An admitted CycleRun never changes.** It keeps its revision and guidance
  snapshots. Only later admissions see the new policy.
- **Apply requires** the expected version, the preview digest, and a non-empty
  rationale. Stale version or digest returns a conflict carrying the latest
  effective policy.
- **Revert goes through the same preview and apply path**, producing a new
  version. It never deletes or rewrites history.
- `CycleRevision` stays the execution revision; the behavior-change record is the
  human-readable audit envelope. They are linked, so a past CycleRun can show
  both the execution snapshot and the change that produced it.

## Acceptance

- Command-module tests cover read, preview, apply, stale-version rejection,
  atomic rollback, immutable guidance replacement, retirement, required
  rationale, and revert-as-new-version.
- CycleRun admission tests prove an admitted run keeps its old snapshots while
  the next run receives the new policy.
- Authorization tests cover same-workspace read, allowed update, cross-workspace
  denial, and delegated agent actor attribution.
- Audit tests prove every applied change — human or agent — carries actor,
  source, rationale, before and after snapshots, version and time.
- **A PostgreSQL-backed test** verifies transaction locking and two-editor
  optimistic concurrency. SQLite cannot prove the live locking behavior.
- Existing Cycle command, revision, memory snapshot, router and authorization
  tests remain green.

## Constraints

- Interface must accept typed adapters later (mandates, overlays) without
  becoming a generic raw-row editor. That is story 30 — a constraint on this
  interface, not a ticket.
- No feature flag, allowlist, or environment gate anywhere in this set.

## Sequencing

Blocks slices 2–6.

redawear:
Closes #782

Slice 1 of 6 of the workspace behavior editor (#738). This is the keystone —
#783, #784, #785, #786 and #787 all build on this module and this audit contract.

## What changed

`brain/systems/cycles/behavior_policy.py` owns read, preview, apply, history and
revert for Cycle policy. `async_update_cycle` now routes through it, and the old
direct-mutation block is **deleted**, so API routes and agent tools share one
path and one audit contract. Nothing mutates policy tables directly any more.

Apply is one transaction boundary. It runs in a savepoint under a
`SELECT ... FOR UPDATE` on the Cycle row, and in that single unit it: writes the
Cycle fields, retires guidance dropped from the active set, adds new immutable
guidance rows, records a `CycleRevision`, writes the `BehaviorChangeAudit` row,
and publishes the existing `cycles_changed` event.

The guarantees the ticket asked for:

- **Guidance text is immutable.** A wording edit retires the old row and creates
  a new one. Retirement changes effective state; it never deletes content.
- **An admitted CycleRun never changes.** It keeps its revision and guidance
  snapshots; only later admissions see the new policy.
- **Apply requires** the expected version, the preview digest, and a non-empty
  rationale. A stale version or digest returns a conflict carrying the latest
  effective policy.
- **Revert reuses the same preview and apply path** and produces a new version,
  linked through `reverted_from_id`. It never rewrites history.

Migration `0059` adds `behavior_change_audits`. Its `has_table` guard has no
explicit schema, matching the idiom `0058` established so the check resolves
through the same `search_path` as the statements it guards.

## The review found a real forward-compatibility bug, and it is fixed here

The cross-family code-quality review (Codex, thermo-nuclear rubric) flagged that
snapshot validation required an **exact** field set. That is right when building
a snapshot from a live Cycle and wrong when reading one back out of the audit
table: the first policy field slices 2–6 add would have made every earlier
`before_snapshot` unreadable, so **revert and history would break on old rows**.
Immutable history you cannot read is not history.

Second commit fixes it. Persisted snapshots carry `snapshot_version`. Decoding a
stored snapshot starts from the current effective policy, overlays the stored
fields it recognises, drops the rest, and validates what survives. A test writes
a snapshot with one field missing and one unknown field present and proves a
revert from it still succeeds.

Three smaller seam fixes rode along: one omission sentinel instead of two,
`publish_cycle_change_safe` / `publish_cycle_change_strict` instead of a boolean
mode flag, and `serialize_behavior_change` moved to `serializers.py` — which
removes the module cycle `memory.py` was working around with a local import.

## Two findings I did NOT fold, filed instead

- **#790** — the snapshot boundary is still a raw dict contract and five slices
  are about to build on it. That is a typed-representation redesign, and it
  should be decided **before #783** starts rather than grown into this PR.
- **#791** — `behavior_change_audits` is named generically but requires a
  non-null `cycle_revision_id`. Which way to resolve that is a design decision,
  not a mechanical fix.

## Verification

- Repo CI command, run serially on an idle machine (load 3.9 at start):
  → **4876 passed, 11 skipped**.
- `alembic heads` prints exactly one head: `0059_behavior_change_audits`.
- The diff is backend only; it touches no `frontend/**` file, so that lane does
  not apply.

**One acceptance line I could not execute:** the PostgreSQL locking and
two-editor concurrency test. It is written
(`tests/test_cycle_behavior_policy_concurrency.py`), correctly marked
`requires_db`, and was collected and skipped because `TEST_DB_URL` is not set
here. SQLite cannot prove live locking, so that check is genuinely unverified on
this machine — it needs a run against real Postgres.

There is no UI yet — that is slice 3. The behavior is visible through the Cycle
edit path that already exists:

1. Edit a Cycle on illo-dev (name, prompt or schedule) and save. It should work
   exactly as before.
2. Check `behavior_change_audits` — that edit should now have a row with your
   actor, the rationale, before and after snapshots, and version 1.
3. Edit the same Cycle's guidance text. The old guidance row should be **retired,
   not deleted**, and a new row should exist.
4. Ask Illo to change the same Cycle. The agent's change must produce a row in
   the same table, with the actor recorded as the agent — same path, same audit.

### DIFF
--- a/brain/app/api/routers/agent_mcp.py
+++ b/brain/app/api/routers/agent_mcp.py
@@ -44,7 +44,7 @@ from brain.systems.cycles.commands import (
     async_update_cycle as command_update_cycle,
     cycle_change_event,
 )
-from brain.systems.cycles.events import publish_cycle_change
+from brain.systems.cycles.events import publish_cycle_change_safe
 from brain.systems.cycles.serializers import (
     serialize_cycle,
     serialize_cycle_guidance,
@@ -1503,7 +1503,7 @@ async def _handle_mcp_request(
             cycle_change = tool_payload.pop("_cycle_change", None)
             await db.commit()
             if isinstance(cycle_change, dict):
-                publish_cycle_change(
+                publish_cycle_change_safe(
                     action=str(cycle_change.get("action") or "update"),
                     **dict(cycle_change.get("event") or {}),
                 )
--- a/brain/app/api/routers/agent_mcp.py
+++ b/brain/app/api/routers/agent_mcp.py
@@ -1203,7 +1203,10 @@ async def _act_manage_cycle(
         )
         await db.flush()
         await db.refresh(cycle)
-        return _cycle_mutation_payload("update", cycle, {"cycle": serialize_cycle(cycle)})
+        return {
+            "cycle": serialize_cycle(cycle),
+            "_mutates_cycle": True,
+        }
 
     if action == "delete":
         await command_delete_cycle(db, cycle)
@@ -1239,11 +1242,11 @@ async def _act_manage_cycle(
             guidance=_required_capability_string(arguments, "guidance", capability="cycle.manage"),
             rationale=rationale,
         )
-        return _cycle_mutation_payload(
-            "update",
-            cycle,
-            {"cycle": serialize_cycle(cycle), "guidance": serialize_cycle_guidance(guidance)},
-        )
+        return {
+            "cycle": serialize_cycle(cycle),
+            "guidance": serialize_cycle_guidance(guidance),
+            "_mutates_cycle": True,
+        }
 
     if action == "add_output_target":
         target = await command_add_cycle_output_target(
--- a/brain/app/api/routers/cycles.py
+++ b/brain/app/api/routers/cycles.py
@@ -22,7 +22,7 @@ from brain.systems.cycles.commands import (
     async_update_cycle as command_update_cycle,
     cycle_change_event,
 )
-from brain.systems.cycles.events import publish_cycle_change
+from brain.systems.cycles.events import publish_cycle_change_safe
 from brain.systems.cycles.common import SCHEDULED_DIGEST_RUN_KIND
 from brain.systems.cycles.serializers import serialize_cycle, serialize_cycle_run
 from brain.systems.cycles.service import async_run_cycle_now
@@ -114,7 +114,7 @@ async def create_cycle(
     payload = serialize_cycle(cycle)
     event = cycle_change_event(cycle)
     await db.commit()
-    publish_cycle_change(
+    publish_cycle_change_safe(
         action="create",
         **event,
     )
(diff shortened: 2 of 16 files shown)


redawear:

redawear:

Issue #783, opened 2026-08-10
[behavior-editor] Slice 2 — API routes over the command module, with 409 conflict semantics


redawear:
> *This was generated by AI during triage.*

Slice **2 of 6** of [#738](https://github.com/Illospace/illospace/issues/738),
sliced on Reda's instruction (2026-08-10). Read the PRD first.

## Scope

HTTP routes over the behavior-policy command module from slice 1. Routes call the
module; they never touch policy tables.

- **Read effective policy** — returns an opaque version, current Cycle
  configuration, active guidance, read-only output targets, source metadata, and
  the latest change summary.
- **Preview** — read-only. Normalizes and validates the proposed policy, computes
  a field-aware diff, reports warnings, states which runs are affected, and
  returns a preview digest tied to the current version.
- **Apply** — requires expected version, preview digest, and a non-empty
  rationale.
- **History** — paginated, ordered by effective version.
- **Revert** — preview and apply, through the same path.

## The conflict contract

Apply returns **HTTP 409** when the version or the digest is stale. The response
**includes the latest effective policy**, so the client can refresh without
losing its draft. That payload shape is what makes slice 4's conflict recovery
possible — get it right here.

## Validation

Reject invalid schedules, empty missions, unsupported models, and duplicate
guidance **before** apply. Broken policy must not become active.

## Authorization

Reads and writes are scoped to the current workspace and existing Cycle
authority. Read access follows existing Cycle read access; apply and revert
follow existing Cycle update authority. One workspace cannot inspect or modify
another. Agent-authored changes carry the delegated actor and the originating run
reference.

## Acceptance

- API tests cover effective-policy shape, semantic diff shape, validation errors,
  409 conflict payloads, history pagination, apply, and revert.
- Cross-workspace access is denied, with a test.
- Policy change events are available to product telemetry and the audit log.
- All API values are timezone-aware UTC.

## Constraints

- No feature flag, allowlist, or environment gate.
- No route mutates a policy table directly.

## Sequencing

Depends on slice 1 (command module). Blocks slices 3–6.

redawear:
> *This was generated by AI during triage.*

**Blocked on #790 by Reda's decision (2026-08-10).**

This slice was going to be written against the raw dictionary snapshot contract
#782 introduced. Reda decided the typed contract lands first, so #783 is not
startable until #790 is done — otherwise the API routes get written against dict
keys and then rewritten.

[...]

PR #793, opened 2026-08-10
refactor(cycles): type the versioned policy snapshot (behavior editor precondition)

### LINKED ISSUE #790: [behavior-editor][refactor] Cycle policy snapshots are an untyped dict contract that five slices are about to build on

redawear:
Closes #790

**Stacked on #792** (issue #782). Base this PR's review on its own two commits;
the parent's work is already reviewed in #792. Merge #792 first, then this.

Reda sequenced this ahead of slice 2 (#783) on 2026-08-10: five slices (#783–#787)
were about to be written against the raw dictionary contract and then rewritten.

## What changed

`CyclePolicySnapshot` is now a typed, versioned value that owns the JSON key
names. Encode writes versioned JSON; decode accepts legacy unversioned
snapshots, overlays the fields it recognises onto the current effective policy,
ignores unknown ones, and validates what survives.

That decode behavior is **#782's**, re-expressed in types rather than
re-litigated. Its compatibility test survives and gained two siblings:
a legacy-unversioned snapshot, and rejection of invalid versions.

- Effective policy, previews, audit records, digests, apply, history and revert
  all carry the typed snapshot. The dictionary validator and the raw replacement
  path are **deleted**.
- All **15** `CyclePolicyPatch` fields lose `Any` and gain real types. One typed
  `UNSET_CYCLE_FIELD` sentinel survives.
- Preview digests are byte-identical to the previous representation, so previews
  taken before this lands stay valid.
- **No migration.** This changes representation, not storage. `alembic heads`
  still prints one head.

## The review found the gap in my own acceptance claim, and it is closed

The ticket's headline is *"adding a policy field requires editing the snapshot
type and nothing else a type checker could not find."* The cross-family review
(Codex, thermo-nuclear rubric) pointed out that the **write** side never met
that: constructor checks catch a missing read or validation mapping, but nothing
catches a missing branch in `apply_to()`. A maintainer could add a field that
appears in previews and audit history and is never written to the live Cycle.

Second commit closes it with a coverage guard driven by
`dataclasses.fields()` — never a hand-written list, which would recreate the
defect. It mutates each field, applies, and asserts the Cycle attribute changed.
Proven to bite: a throwaway unwritten field fails with
`CyclePolicySnapshot.apply_to() did not write snapshot field 'last_status'`.
Guidance is excluded explicitly and in a comment, because the apply command
writes it separately through `_replace_active_guidance()`.

## Known limitation, recorded rather than fixed

`retry_policy` and `guidance` remain a mutable `dict` and `list` inside a frozen
snapshot, so a caller could in principle mutate a snapshot after its digest was
computed. Nothing in the current code does, and immutable containers are a wider
change than this ticket. Flagging it rather than growing the diff.

## Verification

- Repo CI command, run serially on an idle machine (load 3.1 before, 3.7 after):
  → **4882 passed, 11 skipped**. That is 6 more than the parent branch's 4876.
- `alembic heads` → one head, `0059_behavior_change_audits`. No migration added.
- No module outside the snapshot type reads snapshot JSON keys. (The
  `execution_policy_key` literals still in `serializers.py` are the pre-existing
  API response serializers for the `Cycle` and `CycleRevision` ORM rows — a
  different concern, untouched by this PR.)

This is a representation change with no behavior change, so the check is that
nothing moved:

1. Edit a Cycle on illo-dev. It should behave exactly as it does after #792, and
   still write a `behavior_change_audits` row.
2. Revert that change from history. It should succeed and produce a new version.
3. Both should work on a Cycle whose audit rows were written **before** this
   lands — that is the compatibility decode doing its job.

---

## ⚠️ This PR will NOT auto-close #790

Its base is a feature branch, not the default branch, so GitHub refused the
`Closes #790` link — `closingIssuesReferences` is empty and verified empty, not
assumed. When this stack merges, **#790 stays open**.

Two ways to fix it, either is fine:

- After #792 merges, retarget this PR's base to `main`. The closing link then
  registers on its own.
- Or just close #790 by hand after the merge.

### DIFF
--- a/brain/systems/cycles/behavior_policy.py
+++ b/brain/systems/cycles/behavior_policy.py
@@ -9,9 +9,18 @@ from __future__ import annotations
 import hashlib
 import json
 from copy import deepcopy
-from dataclasses import dataclass
+from dataclasses import dataclass, fields, replace
 from datetime import datetime, timezone
-from typing import Any, Mapping
+from typing import (
+    Any,
+    ClassVar,
+    Final,
+    Mapping,
+    TypeAlias,
+    TypeGuard,
+    TypeVar,
+    cast,
+)
 
 from sqlalchemy import func, select
 
@@ -54,6 +63,7 @@ __all__ = [
     "CyclePolicyConflict",
     "CyclePolicyPatch",
     "CyclePolicyPreview",
+    "CyclePolicySnapshot",
     "EffectiveCyclePolicy",
     "UNSET_CYCLE_FIELD",
     "async_apply_cycle_policy_change",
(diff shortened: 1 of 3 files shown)


redawear:

redawear:

Issue #779, opened 2026-08-10
[runtime] Storage policy row — one runtime-editable record for retention window and capacity thresholds


redawear:
> *This was generated by AI during triage.*

Slice **1 of 3** of [#752](https://github.com/Illospace/illospace/issues/752),
sliced on Reda's instruction (2026-08-10). Read #752 first — it holds the
evidence, the three causes, and the prior art. This ticket is the foundation the
other two read from, so it lands first.

## The rule this exists to enforce

> "If this ticket ships a fourth hardcoded window, it has failed." — #752

Every bounding rule in the codebase today is a Python constant or an env var:
`CANVAS_QUIET_HOURS = 24` (`brain/jobs/pipelines/cortex_occupancy.py:20`),
`PROJECT_DRAFT_UNPUBLISHED_RETENTION = timedelta(days=7)`
(`brain/systems/cortex/project_context/draft_lifecycle.py:21`), and every alarm
threshold in `brain/app/scheduler/health_policy.py:9-26`. Changing any of them is
a code change, a review and a deploy. Illo cannot tune one.

## Scope

One runtime-editable record holding:

- the retention window for finished agent workspaces,
- the warn and critical capacity thresholds,
- whether automatic reclamation is allowed.

## Follow the precedent, do not invent one

`CycleGuidance` (`brain/platform/db/models/cycle.py:165-188`) is the shape to
copy: a `rationale` field and an `is_active` flag, edited through a `manage_*`
action, so a change is audited and reversible with no deploy.

Note that `vault_config` (`brain/platform/db/models/vault.py:82-91`) is a generic
store, but its writable-key whitelist is currently exactly one key
(`brain/systems/runtime_settings/preferences.py:106,154`). Either widen it
deliberately or use a purpose-built row — do not quietly add a second generic
key-value store.

## Acceptance

- Illo can read and change the retention window and both thresholds at runtime,
  with a recorded rationale, and no deploy.
- A change is audited and reversible.
- No new hardcoded retention constant is introduced by this slice.
- `CANVAS_QUIET_HOURS` and `PROJECT_DRAFT_UNPUBLISHED_RETENTION` move onto this
  surface, so there is one place where Illo's retention rules live. If that move
  turns out to be larger than this slice, split it — but say so, do not leave two
  homes silently.

## Sequencing

Blocks the capacity sensor and the reclaim lever, which both read this row.

redawear:
Draft PR: #797 (stacked on #794 — merge #794 first).

[...]

PR #794, opened 2026-08-10
chore: consolidate the four open draft PRs into one merge (#755, #773, #782, #790)

### LINKED ISSUE #755: [deploy] Split wait_for_worker_exit — one function now carries two drain policies and three exit statuses

### LINKED ISSUE #773: [scheduler][refactor] Trigger mode tag does not narrow the trigger type — dispatch still needs casts

### LINKED ISSUE #782: [behavior-editor] Slice 1 — behavior-policy command module and the immutable change audit envelope

### LINKED ISSUE #790: [behavior-editor][refactor] Cycle policy snapshots are an untyped dict contract that five slices are about to build on

redawear:
> *This was generated by AI during an autonomous routine run (R3).*

One branch folding the four draft PRs that were open at 20:40Z, per the standing
rule that more than two open PRs become one merge instead of N.

Closes #755
Closes #773
Closes #782
Closes #790

## Why this exists instead of the four PRs

[#793](https://github.com/Illospace/illospace/pull/793) was stacked on
[#792](https://github.com/Illospace/illospace/pull/792). This repo squash-merges,
and a squash-merged parent leaves the child branch with no shared ancestry — the
child then conflicts across every file the parent touched. Folding removes that
trap before you hit it.

It also fixes a second thing: `Closes #790` could not register on #793, because
GitHub refuses closing links on a PR based on a feature branch. On this branch it
registers, so #790 closes on merge instead of needing a manual close.

The four originals stay open. Close them by hand after this merges, or merge them
individually instead and close this one — nothing here depends on which you pick.

## What is in it

| Was | Issue | Change |
|---|---|---|
| #788 | #755 | `wait_for_worker_exit` splits into a covered wait and an idle wait, one drain policy each |
| #789 | #773 | The failure-guard trigger mode tag narrows the trigger type; all six casts are gone |
| #792 | #782 | Behavior-policy command module, one immutable audit row per change, migration `0059` |
| #793 | #790 | The Cycle policy snapshot becomes a typed, versioned record; the untyped dict contract is gone |

Merged with **zero conflicts** — the three source branches share no files except
`behavior_policy.py`, which #792 and #793 already owned as a pair.

## Verification (local — this repo has no CI minutes)

- `alembic heads` → exactly **one** (`0059_behavior_change_audits`).
- Fast suite: **4883 passed, 11 skipped** in 113s, machine load 2.3 (uncontended,
  so the number means something).
- No `frontend/**` file changes, so no frontend lane applies.

## What to watch after you deploy

1. A Cycle policy edit writes exactly one row in `behavior_change_audits`, and the
   Cycle keeps running on the new policy.
2. Reverting that edit restores the prior policy and writes a second audit row —
   it does not mutate the first.
3. A deploy still drains workers normally (the `#755` split is a pure refactor; a
   stuck or skipped drain is the failure to look for).

## Known gap, unchanged from #792

The PostgreSQL two-editor locking test is written and correctly marked
`requires_db`. `TEST_DB_URL` is not set on this machine, so it is collected and
skipped, and SQLite cannot prove live row locking. It needs one run against real
Postgres — that is the one claim here no local run has earned.

### DIFF
--- a/brain/app/api/routers/agent_mcp.py
+++ b/brain/app/api/routers/agent_mcp.py
@@ -44,7 +44,7 @@ from brain.systems.cycles.commands import (
     async_update_cycle as command_update_cycle,
     cycle_change_event,
 )
-from brain.systems.cycles.events import publish_cycle_change
+from brain.systems.cycles.events import publish_cycle_change_safe
 from brain.systems.cycles.serializers import (
     serialize_cycle,
     serialize_cycle_guidance,
@@ -1503,7 +1503,7 @@ async def _handle_mcp_request(
             cycle_change = tool_payload.pop("_cycle_change", None)
             await db.commit()
             if isinstance(cycle_change, dict):
-                publish_cycle_change(
+                publish_cycle_change_safe(
                     action=str(cycle_change.get("action") or "update"),
                     **dict(cycle_change.get("event") or {}),
                 )
--- a/brain/app/api/routers/agent_mcp.py
+++ b/brain/app/api/routers/agent_mcp.py
@@ -1203,7 +1203,10 @@ async def _act_manage_cycle(
         )
         await db.flush()
         await db.refresh(cycle)
-        return _cycle_mutation_payload("update", cycle, {"cycle": serialize_cycle(cycle)})
+        return {
+            "cycle": serialize_cycle(cycle),
+            "_mutates_cycle": True,
+        }
 
     if action == "delete":
         await command_delete_cycle(db, cycle)
@@ -1239,11 +1242,11 @@ async def _act_manage_cycle(
             guidance=_required_capability_string(arguments, "guidance", capability="cycle.manage"),
             rationale=rationale,
         )
-        return _cycle_mutation_payload(
-            "update",
-            cycle,
-            {"cycle": serialize_cycle(cycle), "guidance": serialize_cycle_guidance(guidance)},
-        )
+        return {
+            "cycle": serialize_cycle(cycle),
+            "guidance": serialize_cycle_guidance(guidance),
+            "_mutates_cycle": True,
+        }
 
     if action == "add_output_target":
         target = await command_add_cycle_output_target(
--- a/brain/app/api/routers/cycles.py
+++ b/brain/app/api/routers/cycles.py
@@ -22,7 +22,7 @@ from brain.systems.cycles.commands import (
     async_update_cycle as command_update_cycle,
     cycle_change_event,
 )
-from brain.systems.cycles.events import publish_cycle_change
+from brain.systems.cycles.events import publish_cycle_change_safe
 from brain.systems.cycles.common import SCHEDULED_DIGEST_RUN_KIND
 from brain.systems.cycles.serializers import serialize_cycle, serialize_cycle_run
 from brain.systems.cycles.service import async_run_cycle_now
@@ -114,7 +114,7 @@ async def create_cycle(
     payload = serialize_cycle(cycle)
     event = cycle_change_event(cycle)
     await db.commit()
-    publish_cycle_change(
+    publish_cycle_change_safe(
         action="create",
         **event,
     )
@@ -183,7 +183,7 @@ async def delete_cycle(
     await command_delete_cycle(db, cycle)
     event = cycle_change_event(cycle)
     await db.commit()
-    publish_cycle_change(
+    publish_cycle_change_safe(
         action="delete",
         **event,
     )
@@ -225,7 +225,7 @@ async def run_cycle(
         cycle_id,
         run_kind=SCHEDULED_DIGEST_RUN_KIND,
     )
-    publish_cycle_change(
+    publish_cycle_change_safe(
         action="run",
         **event,
     )
--- a/brain/app/api/routers/cycles.py
+++ b/brain/app/api/routers/cycles.py
@@ -169,12 +169,7 @@ async def update_cycle(
     await db.flush()
     await db.refresh(cycle)
     payload = serialize_cycle(cycle)
-    event = cycle_change_event(cycle)
     await db.commit()
-    publish_cycle_change(
-        action="update",
-        **event,
-    )
     return payload
 
 
--- a/brain/app/scheduler/scheduler_failure_guard.py
+++ b/brain/app/scheduler/scheduler_failure_guard.py
@@ -11,7 +11,6 @@ from typing import (
     Protocol,
     Sequence,
     TypeAlias,
-    cast,
 )
 
 from sqlalchemy import and_, delete, func, or_, select
@@ -31,10 +30,11 @@ from brain.systems.failure_guard.core import (
     FailureGuardLifecycleEvent,
     FailureGuardLatch,
     FailureGuardStatefulTrigger,
+    FailureGuardStatefulTriggerRegistration,
+    FailureGuardStatelessTriggerRegistration,
     FailureGuardTrigger,
     FailureGuardTriggerKind,
     FailureGuardTriggerMode,
-    FailureGuardTriggerRegistration,
     require_failure_guard_registrations,
     FailureGuardTriggerResult,
     FailureGuardTriggerState,
@@ -512,11 +512,11 @@ class SchedulerFailureGuardStatefulTrigger(
     """One state-owning scheduler failure trigger."""
 
 
-SchedulerFailureGuardTriggerImplementation: TypeAlias = (
-    SchedulerFailureGuardTrigger | SchedulerFailureGuardStatefulTrigger
-)
 SchedulerRegisteredFailureGuardTrigger: TypeAlias = (
-    FailureGuardTriggerRegistration[SchedulerFailureGuardLifecycleContext]
+    FailureGuardStatelessTriggerRegistration[SchedulerFailureGuardTrigger]
+    | FailureGuardStatefulTriggerRegistration[
+        SchedulerFailureGuardStatefulTrigger
+    ]
 )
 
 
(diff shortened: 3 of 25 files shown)


redawear:

Superseded by **[#799](https://github.com/Illospace/illospace/pull/799)**, which folds this PR and the rest of the train into one branch.

Why: #797 and #798 were based on #794's branch, and this repo squash-merges. Merging #794 first would have left both children conflicting across every file it touched. #799 removes that trap, and it registers all six `Closes` links — which a feature-branch base refuses, so #779 and #783 would otherwise have stayed open after their merge.

Verified on #799: one alembic head (`0060_storage_policies`), 4902 passed / 11 skipped, machine uncontended.

This PR stays open as a fallback. Close it by hand after #799 merges.

redawear:

Closing as superseded. This consolidation (#755, #773, #782, #790) was folded into #799, which merged to `main` on 2026-08-11. Each constituent PR's content was verified on `main` by content probe before closing.

Issue #796, opened 2026-08-11
[runtime][refactor] Storage policy fields are hand-repeated in nine blocks — give them a typed value object before #780/#781


redawear:
> *This was generated by AI during triage.*

Raised as a NON-BLOCKING finding by the cross-family code-quality review of the
#779 branch (`illo-qa/779-storage-policy-row`). Filed rather than folded because
the remedy rewrites most of the module #779 only touches in two small places.

## Problem

The six storage-policy fields are manually repeated in **nine blocks** inside
`brain/systems/storage_policy.py`:

1. serialization
2. append parameters
3. validation / construction
4. update parameters
5. patch merging
6. revert copying
7. action parameters
8. empty-update detection
9. forwarding

They are then repeated again in the tool schema and the tool handler.

Nothing checks that the eleven agree. A new field can be silently omitted from
revert, from history output, or from update forwarding — and each of those is a
quiet wrong answer rather than an error.

## The second half: consumers read the ORM model

`async_get_storage_policy` returns the database model itself. There are already
five external load sites — two in `brain/jobs/pipelines/cortex_occupancy.py`, one
in `brain/jobs/pipelines/workspace_gc.py`, and two in
`brain/systems/cortex/project_context/draft_lifecycle.py` — and each copies the
load, the field selection, and the hours-to-`timedelta` conversion.

**This is why it matters now.** Slices #780 (capacity sensor) and #781 (reclaim
lever) both read this row. If they land against the ORM model, that is seven load
sites reaching into database columns, and the conversion logic is copied twice
more.

## Fix direction

Introduce a small typed `StoragePolicyValues` (the canonical read shape, with the
`timedelta` conversions done once) and `StoragePolicyPatch` (the canonical write
shape). Let append, update, revert, serialization and every consumer share that
one contract, and derive the tool schema and the empty-update check from it rather
than restating the field list.

This is the same remedy that worked for the Cycle policy snapshot in #790 —
`dataclasses.fields()` as the single source of truth.

## Acceptance

- Adding a storage-policy field touches the value object and at most one other
  hand-edited site; a test proves the rest derive.
- A field that can be updated cannot be missing from revert, history, or the tool
  schema — proven by a test on a throwaway field.
- Consumers receive typed values with the `timedelta` conversion already applied,
  and no consumer imports the ORM model to read policy.
- No behavior change: the same windows and thresholds apply.

## Sequencing

Depends on #779 merging. **Should land before #780 and #781** so those slices read
[...]

Issue #795, opened 2026-08-11
[behavior-editor][refactor] Adding one policy field takes 11 synchronized edits — settle the contract before slice 3


redawear:
> *This was generated by AI during triage.*

Raised by the cross-family code-quality review of the #783 branch
(`illo-qa/783-behavior-policy-api-routes`), as a **BLOCKING** finding. Left out of
that PR because the remedy includes renaming API fields, which is a contract
decision rather than a mechanical fix.

## Problem

Adding one editable **scalar** policy field currently requires **11 synchronized
maintenance sites**, excluding the database and tests:

1. Snapshot declaration
2. `from_cycle`
3. `apply_to`
4. `validated` reconstruction
5. Patch declaration
6. Patch projection
7. API proposal schema
8. Proposal-to-patch adapter
9. Editable-fields list
10. Snapshot serializer
11. Response schema

A field needing a special diff shape adds a twelfth.

## Why it matters now, not later

Nothing type-checks that the eleven agree. A future slice can therefore:

- accept and apply a field while omitting it from history or the 409 body, or
- expose a field in preview that never reaches the live Cycle.

Both are silent. Slices #784–#787 all build on this surface, so every slice
written against the current contract multiplies the cost of fixing it.

This is the same class of defect as #790, one level up. #790 was about the
snapshot **type**; this is about the eleven places the same field must be spelled
across the command module, the API schemas and the serializers. Reda's call on
#790 was to settle the contract *before* slice 2 started, for exactly this reason.

## Fix direction

The reviewer's three-part remedy:

- **Align API and command field names** so no translation layer is needed.
- **Derive** the proposal-to-patch adapter and the editable-fields list rather
  than hand-maintaining them — `dataclasses.fields()` on the typed snapshot is
  the mechanism #790's fix already proved.
- Make **one typed snapshot contract own response serialization**, instead of a
  parallel serializer that must be kept in step.

## Acceptance

- Adding a scalar policy field touches **at most two** hand-edited sites, and a
  test proves the rest derive.
- A field that is accepted and applied cannot be missing from history, the
  preview diff, or the 409 body — proven by a test on a throwaway field, the way
  #790's coverage guard was.
- No route or response shape changes. This is a refactor of how the contract is
  expressed, not of what it returns.

## Sequencing

**Gates slice 3 (#784).** Do this before more slices are written against the
eleven-site contract. Depends on #783 merging first.

Issue #784, opened 2026-08-10
[behavior-editor] Slice 3 — read-only effective-policy view in the Cycles workspace


redawear:
> *This was generated by AI during triage.*

Slice **3 of 6** of [#738](https://github.com/Illospace/illospace/issues/738),
sliced on Reda's instruction (2026-08-10). Read the PRD first.

This slice ships value on its own: it answers "what will Illo do on the next run"
before anything becomes editable.

## Scope

Add an effective-policy view to the selected Cycle detail in the **existing**
Cycles workspace. Read-only.

It shows, in one place:

- the mission prompt, schedule, enabled state, and model and thinking overrides,
- the complete active guidance set,
- output targets, marked read-only,
- for each policy value: its source and last change time.

**Active guidance is visually separate from retired guidance.** Retired guidance
stays readable under History, so earlier behavior can be investigated — but old
prose must never look live. That separation is the point of the whole PRD: today
retired prose can sit in the database looking active.

## The constraint that shapes this

**Do not create a second competing Cycle editor.** This extends the surface that
already exists. If the existing Cycle detail does not have room for it, change
that page — do not open a parallel one.

The UI leads with "Active now." History and prior versions live under History.

## Acceptance

- A workspace member can see a Cycle's effective behavior in one place.
- Active and retired guidance are unmistakably distinct.
- Each value shows where it came from and when it last changed.
- Output targets appear and are clearly not editable in this slice.
- All dates use the user's display timezone while API values stay UTC.
- Existing Cycles page tests remain green.

## Sequencing

Depends on slice 2 (API routes). Blocks slice 4.

redawear:
> *This was generated by AI during triage.*

**Gate: #795 lands before this slice starts.**

The #783 review found that adding one editable scalar policy field currently takes **11 synchronized hand edits** across the command module, the API schemas and the serializers, with nothing type-checking that they agree. Slice 3 is the first slice that would be written against that contract; every slice written against it multiplies the cost of fixing it.

This is the same call Reda made on #790 before slice 2 — settle the contract first. #795 has the fix direction and acceptance.

Issue #781, opened 2026-08-10
[runtime] Reclaim lever — an audited tool that inventories workspaces and reclaims what policy allows


redawear:
> *This was generated by AI during triage.*

Slice **3 of 3** of [#752](https://github.com/Illospace/illospace/issues/752),
sliced on Reda's instruction (2026-08-10). Read #752 first for the evidence and
the prior art.

## Problem

Illo has no owned way to reclaim space. There is no delete or retention verb in
the tool catalog. Deletion is reachable only through raw `exec_command`, whose
blocklist (`brain/systems/runs/tool_catalog/handlers/files.py:350-357`) stops
`rm -rf /` and little else.

So today Illo either does nothing, or does something unaudited.

## Scope

A tool that inventories agent workspaces with their sizes, and reclaims the ones
the storage policy marks reclaimable.

**The safety rules live in code, not in the prompt:**

- never touch a workspace whose parent run is non-terminal,
- never touch a path that is not a workspace,
- always report what was freed.

The manual cleanup on #750 used exactly these rules and they held across 833
directories.

`brain/app/scheduler/overdue_monitor.py:393-420` is the shape for the automatic
path: measure, re-confirm the observation is still current, attempt a bounded
remediation, escalate to a human when attempts run out.

## Relationship to the existing GC

`brain/jobs/pipelines/workspace_gc.py` (from #750) already reclaims finished
headless-worker workspaces on a fixed 48-hour window. This slice does **not**
duplicate it — it gives Illo the inventory-and-reclaim verb, and moves that
48-hour constant onto the policy row so the window stops being a code change.
Check the GC's current behavior before writing a second deletion path; extend it
if that is the honest fit.

## Acceptance

- Automatic reclamation is bounded, refuses to touch workspaces of non-terminal
  runs, and reports bytes freed.
- Illo can inventory workspaces with sizes through a tool, without raw shell.
- `illospace_illo_workspaces` stays bounded across a month with no human action.
- Whether automatic reclamation runs at all is read from the policy row.

## Sequencing

Depends on the storage policy row slice. Best done after the capacity sensor, so
the effect of a reclaim is visible in the trend it writes.

redawear:
> *This was generated by AI during triage.*

**Read [#796](https://github.com/Illospace/illospace/issues/796) before starting this slice.**

[...]

PR #802, opened 2026-08-11
chore: finish the two policy contracts and give meetbot a durable session record (#778, #800, #801)

### LINKED ISSUE #778: [meetings] A meetbot session leaves no durable record — every deploy destroys the only evidence a join was attempted

## Problem

The meetbot service keeps its sessions in memory. Its only trace of a join
attempt — the URL, whether the bot was admitted, what Meet said when it was not —
is the container's stdout.

`docker` keeps those logs only for the current container. Every deploy recreates
`illospace-meetbot-1`, and the evidence goes with it.

Measured on illo-dev, 2026-08-10:

- Container `created` and `started` at `2026-08-10T15:48:50Z`, `RestartCount=0`.
- `docker logs` therefore holds 564 lines, all of them `GET /healthz` since that
  moment. Nothing before the deploy exists.
- `git grep` finds no meeting-session table: the only mention of persistence in
  `brain/systems/meetings/` is a health-observation note in `inbound.py:726`.

## Why it matters

This is not a hypothetical gap; it already cost a diagnosis.

When Reda reported that Illo failed to join a live meeting, there was no log, no
session row and no status record to read. The cause had to be recovered by
querying `agent_run_events` for tool calls — which happened to work only because
the *brain* records its tool calls. Had the request never reached a run, there
would have been no evidence anywhere at all that anyone tried.

The same blind spot produced #704: the bot sat two hours in an empty meeting and
nothing said so.

## Fix direction

Persist one row per meetbot session, written by the brain when it asks and
updated as the session reports. Enough to answer "was it asked, did it get in,
and if not what did Meet say":

- session id, meeting URL, requesting run id, requested-at
- outcome: `requested` / `admitted` / `refused` / `not_admitted` / `left`, with
  the refusal text when Meet gave one
- admitted-at, left-at, observed participant and caption counts

`memory_health_log` is the wrong home — that is a health check log. This is a
session record and deserves its own small table.

The engine already produces the right vocabulary: `_wait_for_admission` returns
`reason="not_admitted"`, and `_raise_if_join_blocked` raises `JOIN_BLOCKED_ERROR`
naming the anonymous-guest block. Today both are only ever logged.

## Acceptance

- A join request writes a row before the browser starts, so an ask is recorded
  even when the service then dies.
- The row's final state distinguishes refused, not admitted, and joined.
- The refusal text from Meet is stored when there is one.
- A redeploy does not lose any of it.
- Someone can answer "did Illo join the 10am meeting, and why not" from the
  database, without shell access to a container.

## Context

Split out of the investigation on the failed live join. Related: #704 (silent
two-hour empty meeting), #658 (meetbot merges never reaching the container).

### LINKED ISSUE #800: [behavior-editor][refactor] The derived policy contract is not guarded at the HTTP boundary, and owns transport concerns

Raised by the cross-family code-quality review of the #795 branch
(`illo-qa/795-policy-field-contract`), which returned **BLOCKING**. #795 was
merged anyway in [#799](https://github.com/Illospace/illospace/pull/799): it
delivers its headline result — adding one scalar policy field went from 11
synchronized sites to **2** — and the findings below are about how the derivation
is expressed, not about behavior. Both the review and a direct measurement
confirmed no route or response change.

Filed rather than folded because the remedy moves code across the command/API
boundary, which is a layering decision rather than a mechanical fix.

## Finding 1 — the coverage guard cannot cross the HTTP boundary

`tests/test_cycle_behavior_policy_api.py::test_throwaway_scalar_reaches_preview_history_and_conflict`
is the test that is supposed to prove a new field cannot go missing. It does not
prove it end to end:

- it builds `behavior_policy.CyclePolicyPatch(throwaway_scalar="after")` directly,
  so the field never passes through `CyclePolicyProposal` or
  `cycles_router._policy_patch`;
- it calls `cycles_router.preview_cycle_behavior_policy(...)` and
  `list_cycle_behavior_policy_history(...)` as plain functions, so FastAPI's
  `response_model` filtering never runs.

The reason it cannot be fixed by monkeypatching harder: `CyclePolicyConfigurationRead`
is built **at import time** by `_cycle_policy_configuration_read_model()` in
`brain/app/api/schemas/cycles.py`, from the real `CyclePolicySnapshot`. A test that
swaps the snapshot class afterwards cannot rebuild that model, so under a real
request the throwaway field would be filtered straight back out.

**This is a gap in the proof, not in production behavior.** A real new field is
declared before import, so it flows into the generated model normally. But the
acceptance criterion #795 was written against — *"a field that is accepted and
applied cannot be missing from history, the preview diff, or the 409 body — proven
by a test"* — is not actually guarded at the surface where the field would be lost.

## Finding 2 — transport policy now lives in the command module

`brain/systems/cycles/behavior_policy.py` declares `_API_EDITABLE` and
`_API_RESPONSE_TYPE` field metadata, and owns `response_payload()` including the
`schedule_human` REST field. `brain/app/api/schemas/cycles.py` then reflects back
over that command type to generate its response model.

So command orchestration, persistence, editor exposure and HTTP representation now
sit in one file, pointing at each other in both directions. The file did get
smaller (972 → 959 lines), but its central contract is less direct.

## Fix direction

- Exercise the throwaway field through a real ASGI request, or rebuild the
  response model inside the test so `response_model` filtering is actually
  applied. Either one closes the hole; asserting on route return values does not.
- Extract the snapshot, its field metadata and the response projection into a
  focused policy-contract module. Leave command orchestration in
  `behavior_policy.py` and let the API schemas consume the contract, so the
  reflection points one way.

## Acceptance

- Adding a scalar policy field is still at most two hand-edited sites.
- A throwaway field is proven to survive **a real HTTP preview, history and 409
  response** — a test that would fail if `response_model` filtered it out.
- `behavior_policy.py` no longer declares API-transport concerns.
- No route or response shape change.

## Sequencing

Should land before more of #784–#787 are written against the contract, for the
same reason #795 itself was done before slice 3: every slice written against the
current expression multiplies the cost of changing it.

### LINKED ISSUE #801: [runtime][refactor] Storage policy patch still restates the field list, and the tool handler synthesizes its signature

Raised by the cross-family code-quality review of the #796 branch
(`illo-qa/796-storage-policy-values`), which returned **BLOCKING**. #796 was merged
anyway in [#799](https://github.com/Illospace/illospace/pull/799): it delivers the
result that gates #780 and #781 — the ORM leak is gone, all five consumers now
receive typed values with the `timedelta` conversion done once, and the review
confirmed no behavior change. The findings below are what it did not finish.

## Finding 1 — adding a field still takes three sites, not two

#796's acceptance said *"adding a storage-policy field touches the value object and
at most one other hand-edited site."* It currently takes three:

1. `brain/platform/db/models/storage_policy.py` — the ORM column
2. `brain/systems/storage_policy.py` — `StoragePolicyValues` (field, storage
   mapping, type, tool schema)
3. `brain/systems/storage_policy.py` — `StoragePolicyPatch`, which restates the
   same field list a second time

`tests/test_storage_policy.py::test_dataclass_fields_drive_patch_storage_and_tool_schema`
does catch the drift, so a forgotten field fails loudly rather than silently. But
detecting duplication is not removing it, and the ticket asked for removal.

**Fix direction:** derive the patch keys from the value contract — the patch is
"the value object, every field optional" and should be generated from it, leaving
the value object and the ORM declaration as the only hand-edited sites.

## Finding 2 — the storage-policy tool handler synthesizes its own signature

`brain/systems/runs/tool_catalog/handlers/composition.py` declares
`_manage_storage_policy(action=..., policy_id=..., rationale=..., limit=...,
**storage_values)` and then assigns a hand-built `inspect.Signature` to its
`__signature__`, generated from `storage_policy_field_schema`.

Introspection therefore advertises named parameters the function does not declare,
while the real call boundary still accepts arbitrary keys. This was introduced by
#796 — it is not pre-existing. It is the mechanism that makes the tool schema
derive, so it earns something; it is also reflection magic sitting on a public
tool boundary, and it is the wrong layer for it.

**Fix direction:** one canonical tool-binding adapter that validates arguments
against the derived schema and builds the patch, used by any tool that exposes a
typed contract — instead of each handler mutating its own reported signature.

## Finding 3 (non-blocking) — the module should now be decomposed

`brain/systems/storage_policy.py` grew from 306 to 446 lines. It is well under the
1000-line threshold, so this is not urgent, but it now holds three separable
things: the typed contract and its codecs, the schema derivation, and the ORM
queries plus revision orchestration.

## Acceptance

- Adding a storage-policy field touches the value object and the ORM model only;
  the patch shape and the tool schema derive, proven by a test.
- No handler assigns its own `__signature__`; the tool boundary validates against
  the derived schema in one shared place.
- No behavior change: the same windows and thresholds apply.

## Sequencing

Should land before or alongside #780 (capacity sensor) and #781 (reclaim lever),
which are the first consumers written against this contract.

redawear:

Three tickets, one merge. This branch is **stacked on
`illo-qa/consolidated-20260811`** (PR #799) — merge that one first, then this
one. Two clicks total.

Closes #778
Closes #800
Closes #801

## What is in here

**#778 — a meetbot join now leaves a durable record.**
The meetbot service kept sessions in memory, so the only trace of a join attempt
was container stdout, which every deploy destroys. Adds a `meetbot_sessions`
table (migration `0061`). The brain writes a `requested` row **before** the
browser starts, so an ask is recorded even if the service then dies. Callbacks
persist `admitted` / `refused` / `not_admitted` / `left`, plus the refusal text
Meet gave and the observed participant and caption counts. meetbot still has no
direct database connection.

**#801 — adding a storage-policy field now takes two edits, not three.**
`StoragePolicyPatch` is derived from `StoragePolicyValues` instead of restating
its field list. The tool handler no longer assigns its own `inspect.Signature`;
a shared `bind_derived_tool_contract()` adapter validates arguments against the
derived schema. Sequenced ahead of #780 and #781.

**#800 — the cycle policy contract is guarded where it can actually fail.**
The snapshot, its field metadata and the HTTP response projection moved out of
`behavior_policy.py` into `behavior_policy_contract.py`, so the reflection
between the command module and the API schemas points one way. The
throwaway-field guard now runs through **real ASGI requests**, so FastAPI's
`response_model` filtering actually executes — previously it called route
functions directly and could not have caught a field lost in transport.
Sequenced ahead of #784–#787.

## Why this shape

#801 and #800 are the residue of the cross-family code-quality reviews that
returned BLOCKING on #796 and #795. Those two shipped in #799 on their headline
results; this branch finishes them, ahead of the slices that build on both
contracts.

## Code review found three real defects. They are fixed, not noted.

Two cross-family (Codex) maintainability reviews both returned **BLOCKING**. The
findings were checked against the diff before acting on them — all confirmed
introduced by this change, not pre-existing. Full verdicts:
`state/evidence/778-thermo-review-20260811T1410.md` and
`state/evidence/800-801-thermo-review-20260811T1410.md`.

The three that mattered:

1. **A failed audit-row write killed the meeting join.** The `requested` row was
   written before `client.join()`, and a `SQLAlchemyError` returned an error
   instead of joining — so the observability table could stop the meeting it
   exists to observe. Now best-effort: it still commits before the browser
   starts on the happy path, but a failure logs and the join proceeds. Guarded by
   `test_join_continues_when_request_record_write_fails`.
2. **Active-status webhooks were awaited inside the session state machine**
   (3 attempts × ~15s + backoff), so a brain outage stalled lobby, admission and
   caption handling. Now delivered on tracked background tasks, drained on
   shutdown.
3. **A second refusal vocabulary** (`JoinRefusedError` / `"refused"`) sat beside
   the engine's existing `not_admitted`, and an unknown value silently became
   `not_admitted` — a wrong answer that looks real. Now one validated outcome
   type; unknown values are rejected explicitly
   (`test_unknown_terminal_outcome_is_rejected_explicitly`).

**And one defect the review did not catch, found while verifying its fix:** the
fix introduced `from meetbot.models import ...` into `brain/`. `api.Dockerfile`
copies only `brain/` and `meetbot.Dockerfile` copies only `meetbot/` — they are
separate deployment units, and before this branch `brain/` imported `meetbot`
zero times. That import would have raised `ModuleNotFoundError` on **every brain
API boot**, while passing every local test, because running from the repo root
puts both packages on `sys.path`. The enum now lives in `brain/`, meetbot keeps
its own, and they meet over the wire. Two tests hold the line:
`test_meetbot_outcomes_match_brain_persistence_contract` (the two sets agree) and
`test_brain_modules_do_not_import_meetbot` (the boundary itself).

## Verification

Run locally on a quiet machine — 10 cores, load 5.1, **zero** competing Codex
processes. This repo is out of CI minutes, so local runs plus staging use are the
evidence.

- **Backend fast suite: 4915 passed, 11 skipped, 87.9s.** The repo's own CI
  command: `pytest tests/ -m "not requires_db and not requires_browser and not live_provider"`.
- **Measured against the parent branch, not asserted.** The same command on
  `illo-qa/consolidated-20260811` gives **4906 passed**. Diffing collected test
  names: **11 added, 2 removed** (net +9).
- The 2 removed are `test_snapshot_contract_derives_api_field_plumbing` and
  `test_throwaway_scalar_reaches_preview_history_and_conflict`. They were
  **replaced**, not dropped: the review showed they exercised a test-only router,
  so they could pass while production wiring was broken. The replacements
  (`test_production_response_model_matches_snapshot_contract`,
  `test_production_router_preserves_policy_field_across_workflow`) drive the real
  router.
- **`alembic heads` prints exactly one head:** `0061_meetbot_sessions`.
- **Brain imports cleanly without meetbot** — verified by importing
  `brain.platform.db.models` from outside the repo root and asserting no
  `meetbot` module loaded.
- The diff touches no `frontend/**` file, so the frontend lane does not apply.
- Every branch merged into this one with **zero conflicts**.

## How to check it on illo-dev after merge

1. **#778 — ask Illo to join a Meet link from Slack**, then:
   `select id, meeting_url, outcome, refusal_text, requested_at, admitted_at from meetbot_sessions order by requested_at desc limit 5;`
   A row must exist even if the bot never gets in, and `outcome` must
   distinguish refused from not-admitted. This is the gap that made #777 take
   three sessions of database archaeology to diagnose.
2. **#801 — run the `manage_storage_policy` tool** with `action="get"`, then an
   update such as `canvas_quiet_hours=12`. Accepted arguments and the returned
   schema must be identical to today; an unknown field must be rejected by name.
3. **#800 — open a Cycle's behavior policy** in the Cycles workspace, preview an
   edit, and apply it. Effective policy, the diff and history must render exactly
   as before — this pair is a refactor and a test fix, with no intended
   behavior change.

## One open disagreement, left for you rather than settled quietly

`StoragePolicyPatch` is now generated at import time by
`dataclasses.make_dataclass`, so its fields are invisible to static analysis,
IDE navigation and `grep` — someone looking for them finds a factory function.

**The two cross-family reviews contradict each other on this.** The review of
#796 said the duplicated field list must be *removed*, not merely drift-tested —
that is what #801 was filed to do, and its acceptance criterion says so. The
review of #801 now says restore the explicit dataclass, because the drift test
was already enough and reflection costs more than the six lines it saved.

Both are defensible, and it is a taste call about this codebase, not a bug. I did
not flip it back on my own authority: reverting would contradict the ticket's
stated acceptance. The rest of #801's finding *was* folded — the service now
accepts one write shape, and the single-caller binder is deleted. If you prefer
the explicit dataclass, say so and it is a small follow-up.


redawear:
Conflicts resolved. #799 was squash-merged and its branch deleted, so this branch shared no ancestry with its own content on `main` — every file #799 touched conflicted even though nothing actually disagreed.

Verified before resolving, rather than assuming:
- `git diff <old-799-tip> origin/main` is **empty** — main's tree is content-identical to the branch this was stacked on.
- `git diff <old-799-tip> HEAD` is **byte-identical** to `git diff origin/main HEAD`, so `main` holds nothing this branch lacks.

So the fix was ancestry, not content: an `-s ours` merge of `origin/main` that records the parent and keeps this tree. The tree hash is **unchanged** (`071267c9`) — it is the exact tree that passed 4915 / 11 skipped, so that gate still stands and did not need re-running. `alembic heads` still prints one head against the new main.

Now MERGEABLE, and `Closes` is registered for #778, #800 and #801 — the retarget to `main` means merging this will close all three.

PR #803, opened 2026-08-11
feat(cycles): read-only effective-policy view in the Cycles workspace (#784)

redawear:
> *Autonomous routine run (R3).*

Behavior-editor **slice 3 of 6** of the [#738](https://github.com/Illospace/illospace/issues/738) PRD. Slice 1 (#782), the typed snapshot (#790) and slice 2's API routes (#783) all merged this morning in #799, which unblocked this.

Closes #784

## What you are looking at

An **"Active now"** panel inside the existing Cycle detail — not a new page and not a second editor. It answers "what will Illo do on the next run" before anything becomes editable.

## I verified it in a browser, here is what it shows

- **Mission and settings** — prompt, schedule rule and human schedule, timezone, enabled, concurrency, retry policy, model and thinking overrides. Every value carries its source (`From api:/cycles/901/behavior-policy`) and its last-change time.
- **Active guidance** — marked `LIVE`, with a green rail and an "N active" count.
- **Output targets** — shown with a `READ ONLY` badge, since this slice does not edit.
- **History** — collapsed by default. Opening it shows a `Historical only — Text below does not guide the next run.` banner, and each retired line is labelled `RETIRED GUIDANCE · NOT ACTIVE`.

That last point is the whole reason the PRD exists: retired prose used to sit in the database looking live. It now cannot be mistaken for active guidance.

- All dates render in the user's display timezone; API values stay UTC.
- Policy fields are read through typed client contracts, so a maintainer adding a field does not hand-edit string-keyed maps.

## How to look at it yourself

```
cd frontend && npm ci && npm run dev
```

then open `http://localhost:5173/cycles?preview=1` — the preview route carries sample policy data, so no backend is needed.

## Evidence

- `npm test` — 254 passed, 0 failed.
- `npm run check` (`svelte-check`) — 6,702 files, **0 errors, 0 warnings**.
- `npm run build` — passed.
- Browser check: view rendered, History expanded, **no JavaScript errors**. The console 500s are `/api/me` and `/api/cortex/*` because no brain API runs in the preview; no behavior-policy request fails.

The diff touches `frontend/**` only, so this is the only CI lane it selects.

PR #806, opened 2026-08-11
fix(cycles): make the behavior-change audit envelope honestly Cycle-only (#791)

### LINKED ISSUE #791: [behavior-editor] The behavior-change audit envelope is named generic but is structurally Cycle-only

redawear:

You answered this on 2026-08-11: **option 1**, make it honest rather than adapter-neutral. No second policy kind is planned, so neutrality would have been speculative.

Closes #791

## What was wrong

The audit envelope advertised itself as generic — it carried `policy_kind`, `target_type` and `target_id` — while `cycle_revision_id` was NOT NULL with a foreign key to `cycle_revisions`. A non-Cycle policy could never have written a row. The columns described a capability the schema does not have.

## What changed

`behavior_change_audits` → `cycle_behavior_change_audits`, `BehaviorChangeAudit` → `CycleBehaviorChangeAudit`, and the two discriminator columns are gone. The unique key is now `(target_id, version)`.

## The migration is the risky part, so here is exactly what it does

The table is **already live** — illo-dev is at `0061` and holds this table — so `0059` is untouched and migration `0062` renames in place off `0061_meetbot_sessions`. One alembic head.

It handles both paths: a fresh install, where the public baseline already created the new name (it drops the empty historical table, and **refuses** if that table has rows), and the deployed database, where the old name holds data and everything is renamed rather than copied.

**The review caught the thing I would have missed.** `ALTER TABLE ... RENAME TO` does not rename implicitly-named constraints. I checked `pg_constraint` on illo-dev and found the three the first cut left behind:

```
behavior_change_audits_pkey
behavior_change_audits_cycle_revision_id_fkey
behavior_change_audits_reverted_from_id_fkey
```

Without the fix, an upgraded database and a fresh install stop converging, and a later migration referencing a constraint by name would have to know which history the database came from. All three are now renamed on upgrade and reversed on downgrade, each guarded by an existence check so replay and fresh installs stay safe.

Also folded from the review: the neither-table state now raises instead of silently marking the migration applied, and a redundant index that duplicated the unique constraint's backing index is gone.

The migration tests now assert the **emitted SQL** rather than mock calls — a mock happily accepts a rename that never happens. They failed 7 of 10 against the unfixed migration.

## Left for a follow-up, on purpose

The exported `BehaviorChangeRecord` dataclass and its serializer helpers keep generic names, so the Cycle-only vocabulary is not yet consistent end to end. That is a separate rename with its own blast radius. Say the word and I will file it.

## Evidence

- Full backend lane on this branch: **4922 passed, 11 skipped, 0 failed** in 99s, run serially at load 3.8 with no Codex worker on this repo.
- `alembic heads`: exactly one, `0062_cycle_behavior_change_audits`.

## What to exercise after merge

The table has no writer yet — slice 3 (#803) is read-only and apply lands in slice 4 — so the check is the deploy itself: confirm the migration runs clean on staging and that `\d cycle_behavior_change_audits` shows every constraint and index under the new name.

### DIFF
--- a/brain/app/api/schemas/cycles.py
+++ b/brain/app/api/schemas/cycles.py
@@ -227,8 +227,6 @@ class CyclePolicyChangeSummaryRead(BaseModel):
 
 class EffectiveCyclePolicyRead(BaseModel):
     workspace_id: str
-    policy_kind: str
-    target_type: str
     target_id: str
     version: int
     revision_id: int | None = None
@@ -275,8 +273,6 @@ class CyclePolicyPreviewRead(BaseModel):
 
 class CyclePolicyChangeRead(CyclePolicyChangeSummaryRead):
     workspace_id: str
-    policy_kind: str
-    target_type: str
     target_id: str
     before_snapshot: CyclePolicySnapshotRead
     after_snapshot: CyclePolicySnapshotRead
--- a/brain/platform/db/alembic/versions/0062_cycle_behavior_change_audits.py
+++ b/brain/platform/db/alembic/versions/0062_cycle_behavior_change_audits.py
@@ -1,0 +1,150 @@
+"""Make the behavior-change audit envelope explicitly Cycle-only.
+
+Revision ID: 0062_cycle_behavior_change_audits
+Revises: 0061_meetbot_sessions
+Create Date: 2026-08-11
+"""
+
+from __future__ import annotations
+
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = "0062_cycle_behavior_change_audits"
+down_revision = "0061_meetbot_sessions"
+branch_labels = None
+depends_on = None
+
+_OLD_TABLE = "behavior_change_audits"
+_TABLE = "cycle_behavior_change_audits"
+
+_OLD_TARGET_VERSION = "uq_behavior_change_audits_target_version"
+_TARGET_VERSION = "uq_cycle_behavior_change_audits_target_version"
+_OLD_CYCLE_REVISION = "uq_behavior_change_audits_cycle_revision"
+_CYCLE_REVISION = "uq_cycle_behavior_change_audits_cycle_revision"
+_OLD_WORKSPACE_APPLIED = "ix_behavior_change_audits_workspace_applied"
+_WORKSPACE_APPLIED = "ix_cycle_behavior_change_audits_workspace_applied"
+_OLD_TARGET_HISTORY = "ix_behavior_change_audits_target_history"
+_TARGET_HISTORY = "ix_cycle_behavior_change_audits_target_history"
+
+
+def _has_table(table_name: str) -> bool:
+    return sa.inspect(op.get_bind()).has_table(table_name)
+
+
+def upgrade() -> None:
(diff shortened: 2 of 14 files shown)


redawear:
Superseded by #810, which folds this branch and the other six open R3 pull requests into one branch cut from current `main`.

Reason: four of the seven PRs targeted a feature branch, so their `Closes` links never registered — merging the trains would have left four issues open for Reda to close by hand. #810 targets `main` and carries all seven links.

All three CI lanes were re-run on the merged tree, serially: backend 4982 passed, frontend 278/278 with `svelte-check` clean, `alembic upgrade head` clean with exactly one head, and 86 database tests. The five merges were conflict-free.

I will close this PR once #810 has merged and I have confirmed the content is on `main`.

redawear:
Superseded and shipped. This PR's content is on `main` via the consolidated [#810](https://github.com/Illospace/illospace/pull/810), merged 2026-08-12 (`7b1b5137`). Verified by a content probe on `origin/main` for the files this branch adds, not by merge status — a stacked PR reports `MERGED` against its parent branch, which answers the wrong question. Closing this one; nothing is lost.

Issue #738, opened 2026-08-08
[frontend][prd] Workspace behavior editor — Cycle policy first


redawear:
## Problem Statement

Illo's effective behavior is stored in runtime data, but people cannot inspect or change that behavior safely in the product. Cycle missions, active guidance, schedules, model choices, output targets, mandates, runtime overlays, skills, and monitored-channel policy can all change what Illo does. Today, a person must write SQL or ask an agent to perform the edit.

This hides the most important fact: which policy is active now. Retired prose can remain active in the database, authorship is hard to see, and a change can have no useful diff or rollback path. The workspace must become the place where people read and control durable behavior, while Slack remains the place for conversation.

## Solution

Build the first vertical slice as a **Cycle Policy Editor** inside the existing Cycles workspace.

The editor shows the effective policy for one Cycle: mission prompt, schedule, enabled state, model and thinking overrides, and the complete active guidance set. A person can prepare a change, review a semantic diff, provide a required rationale, and apply it. Apply is live for future CycleRuns. A CycleRun that is already admitted keeps its immutable revision and guidance snapshots, so an edit cannot change work in progress.

There is no separate approval queue in the first slice. An authorized user explicitly reviews the diff and selects Apply. Revert is the rollback: it creates a new audited change that restores an earlier effective policy. It never deletes or rewrites history.

The first slice does not make every runtime policy editable. Cycle policy comes first because the repository already has CycleRevision, CycleGuidance, scoped Cycle authorization, and per-run snapshots. Mandates and other overlays follow only after they have typed adapters and validation. This avoids a generic raw-data editor.

## User Stories

1. As a workspace member, I want to see a Cycle's effective behavior in one place, so that I know what Illo will do on the next run.
2. As a workspace member, I want active guidance to be visually separate from retired guidance, so that old prose cannot look live.
3. As a workspace member, I want to see the mission prompt, schedule, enabled state, and model overrides together, so that hidden settings do not change my interpretation.
4. As a workspace member, I want each policy value to show its source and last change time, so that I can understand where it came from.
5. As a workspace member, I want to edit the complete active guidance set, so that I can add, replace, or retire guidance deliberately.
6. As a workspace member, I want retired guidance to remain readable in history, so that I can investigate earlier behavior.
7. As an authorized editor, I want a semantic before-and-after diff before apply, so that I can catch an accidental behavior change.
8. As an authorized editor, I want schedule changes shown in both stored and human-readable form, so that cron syntax is not the only evidence.
[...]

PR #807, opened 2026-08-11
feat(cycles): behavior editor slice 4 — edit, review diff and explicit apply

### LINKED ISSUE #785: [behavior-editor] Slice 4 — the edit, review-diff and explicit apply flow

Slice **4 of 6** of [#738](https://github.com/Illospace/illospace/issues/738),
sliced on Reda's instruction (2026-08-10). Read the PRD first.

This is the slice that makes behavior editable. Everything before it was reading.

## Scope

A deliberate edit flow on the effective-policy view from slice 3:

- prepare a change to the mission prompt, schedule and timezone, enabled state,
  model and thinking overrides, and the active guidance set,
- review a **semantic** before-and-after diff,
- supply a required rationale,
- apply explicitly.

## The rules that make it safe

- **Apply is explicit and never automatic.** Typing in a form does not change
  live behavior.
- **The editor never auto-saves.** Leaving with a dirty draft asks for
  confirmation.
- **Apply is disabled without a rationale.** Future teammates need to know why
  behavior changed.
- **The preview states that active runs are unchanged**, so the editor
  understands when the change takes effect.
- **A stale preview is rejected** when another person changed the Cycle. The 409
  from slice 2 carries the latest effective policy — use it to let the person
  review and retry **without losing their draft**.
- **Enabling or disabling a Cycle goes through this same audited flow.** Schedule
  control must not become an unaudited side path.
- **Existing output targets are preserved** even though they are not editable
  here. A save must never silently remove delivery.
- **Revert** is available and produces a new diff, rationale, author and version.
  Rollback never erases the incident.

## Presentation

The review step uses a restrained field diff, not raw JSON. Schedule changes are
shown in **both** stored and human-readable form, so cron syntax is never the
only evidence.

## Acceptance

- UI domain tests cover draft hydration, field-aware diff rendering, active
  versus retired guidance, conflict recovery, and dirty-draft protection.
- UI integration tests cover review-before-apply, disabled apply without
  rationale, successful refresh after apply, and revert confirmation.
- A successful apply refreshes effective policy and history.
- Invalid schedules, empty missions, unsupported models and duplicate guidance
  are refused before apply.

## Sequencing

Depends on slice 3. Blocks slice 5.

redawear:
Slice **4 of 6** of the workspace behavior editor ([#738](https://github.com/Illospace/illospace/issues/738)). This is the slice that makes behavior editable — everything before it was reading.

Stacked on [#803](https://github.com/Illospace/illospace/pull/803) (slice 3), because it edits the same Cycles surface. **Merge #803 first**, then this.

Closes #785

## What you can see in 6 clicks — no backend needed

```
cd frontend && npm run dev
```

Open **http://localhost:5173/cycles?preview=1**

1. **Edit behavior** on the first Cycle → the draft editor opens, headed *"Nothing changes until you review and apply."*
2. Change the mission prompt → **Review change** → a restrained **field diff**: `Mission prompt` with BEFORE and AFTER side by side. Not raw JSON.
3. **Apply change is disabled**, and the panel says *"Active runs are unchanged. Future runs use this policy only after apply."*
4. Type a rationale → apply enables → apply → **Version 2 becomes Version 3**, "Active now" shows the new mission, and History goes from 1 change to 2.
5. **Revert** on a history entry → its own BEFORE/AFTER diff, its own required rationale, and **Apply revert**. A rollback is a new audited version; it never erases the incident.
6. Try to leave with an unsaved draft → the browser asks you to confirm.

I drove all six myself in a browser and they behave as written above.

## The bug that only a browser could find

The preview flow threw `Failed to execute 'structuredClone' on 'Window'` the moment you clicked **Review change** — the primary action of this slice, dead on click one, while 267 tests, `svelte-check` and the build were all green.

The cause: the new preview fixture wrote its own `clone()` on top of `structuredClone`, which cannot clone Svelte's reactive `$state` proxies. This repo already had the answer — `cloneForPostMessage`, and a test at `postMessageClone.test.js:21` that asserts this exact `DataCloneError`. The fix extracts a shared `clonePlainData` that both callers now use, plus a regression test confirmed to fail without it.

## Also in here: a maintainability pass

A code-quality review came back BLOCKING on four findings and all four were verified before folding. Slice 4 had first landed as one file growing **767 → 1,596 lines**, with slices 5 and 6 due to land on it next:

- **Split into real components** — `EffectiveCyclePolicyView` (342), `ActiveCyclePolicyView` (427), `CyclePolicyDraftEditor` (432), `CyclePolicyReview` (343), `CyclePolicyHistory` (284), plus an `effectivePolicyWorkflow.ts` controller (398). Slice 5 now has a `CyclePolicyHistory` to build on instead of a monolith to edit.
- **One `CyclePolicyWorkflowState` union** (view / edit / reviewing / review / applying / conflicted / reverting) replaces a `stage` string plus three loose booleans that could express impossible states.
- **One `POLICY_FIELD_SCHEMA`** is now the single source of editable-field truth; the draft shape, dirty check, proposal mapping and editor controls all derive from it instead of restating it six times.
- **Tests no longer read source text.** Two tests did `readFileSync` on the `.svelte` file and regex-matched it — they asserted a string existed, not that the behavior held, and any extraction would have failed them for no reason. They are replaced by behavior tests, including one that proves a 409 conflict preserves the draft.

## Verification

All run by me, serially, with no Codex worker live (load average 3.9):

- `npm test` — **268 passed, 0 failed**
- `npm run check` — **0 errors, 0 warnings** across 6,874 files
- `npm run build` — succeeded

Frontend-only diff, so this is the only CI lane the change selects. Console during the browser walkthrough showed no app errors — only Vite's HMR websocket (environmental) and Chrome blocking the `beforeunload` confirm, which is the dirty-draft guard firing under a scripted navigation.

## Two review notes I deliberately did not fix

Both are real and both are wider than this ticket. Naming them rather than expanding the diff:

1. `routes/cycles/+page.svelte` and `ThreadCyclesPane.svelte` each duplicate the dirty flag, discard confirmation, apply callback and refresh coordination. A future change to the discard rule has to be made twice.
2. `client.ts` types diff `kind` and `field` as plain strings although rendering branches on specific semantic variants. A new diff kind would silently fall through to generic rendering; literal unions with exhaustive handling would catch it at compile time.

## Assumptions

- A configured model stays valid even when it is absent from the latest model catalog.
- One-time schedules use the server-supported `at:` timestamp form.

### DIFF
--- a/frontend/src/lib/api/client.ts
+++ b/frontend/src/lib/api/client.ts
@@ -518,6 +518,74 @@ export interface CyclePolicyHistoryRead {
   };
 }
 
+/** Mirrors the editable fields in the server CyclePolicyProposal schema. */
+export interface CyclePolicyProposal {
+  prompt?: string | null;
+  schedule_expr?: string | null;
+  timezone?: string | null;
+  enabled?: boolean | null;
+  model_override?: string | null;
+  thinking_override?: string | null;
+  guidance?: string[] | null;
+}
+
+export interface CyclePolicyPreviewRequest {
+  proposal: CyclePolicyProposal;
+}
+
+export interface CyclePolicyApplyRequest extends CyclePolicyPreviewRequest {
+  expected_version: number;
+  preview_digest: string;
+  rationale: string;
+}
+
+export interface CyclePolicyRevertApplyRequest {
+  expected_version: number;
+  preview_digest: string;
+  rationale: string;
+}
+
+export interface CyclePolicyDiffEntryRead {
+  field: string;
+  kind: string;
+  before: CyclePolicyJsonValue;
+  after: CyclePolicyJsonValue;
+  added: string[] | null;
+  removed: string[] | null;
+}
+
+export interface CyclePolicyWarningRead {
+  code: string;
+  message: string;
+}
+
+export interface CyclePolicyAffectedRunsRead {
+  admitted_runs: string;
+  future_runs: string;
+}
+
+export interface CyclePolicyPreviewRead {
+  expected_version: number;
+  preview_digest: string;
+  before: CyclePolicySnapshotRead;
+  after: CyclePolicySnapshotRead;
+  changed_fields: string[];
+  diff: CyclePolicyDiffEntryRead[];
+  warnings: CyclePolicyWarningRead[];
+  affected_runs: CyclePolicyAffectedRunsRead;
+  reverted_from_id: number | null;
+}
+
+export interface CyclePolicyApplyRead {
+  effective_policy: EffectiveCyclePolicyRead;
+  change: CyclePolicyChangeRead;
+}
+
+export interface CyclePolicyConflictDetail {
+  reason: string;
+  latest_effective_policy: EffectiveCyclePolicyRead;
+}
+
 export interface DomainFieldRead {
   id: number;
   domain_id: number;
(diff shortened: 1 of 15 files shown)


redawear:
Superseded by #810, which folds this branch and the other six open R3 pull requests into one branch cut from current `main`.

Reason: four of the seven PRs targeted a feature branch, so their `Closes` links never registered — merging the trains would have left four issues open for Reda to close by hand. #810 targets `main` and carries all seven links.

All three CI lanes were re-run on the merged tree, serially: backend 4982 passed, frontend 278/278 with `svelte-check` clean, `alembic upgrade head` clean with exactly one head, and 86 database tests. The five merges were conflict-free.

I will close this PR once #810 has merged and I have confirmed the content is on `main`.

redawear:
Superseded and shipped. This PR's content is on `main` via the consolidated [#810](https://github.com/Illospace/illospace/pull/810), merged 2026-08-12 (`7b1b5137`). Verified by a content probe on `origin/main` for the files this branch adds, not by merge status — a stacked PR reports `MERGED` against its parent branch, which answers the wrong question. Closing this one; nothing is lost.

Issue #813, opened 2026-08-13
[refactor] The "thread-discussion:" literal has three owners under two different names


redawear:
> *This was generated by AI during triage.*

## What

The string literal `"thread-discussion:"` is defined in **three** places under **two** different constant names:

| file | constant | concern |
|---|---|---|
| `brain/contracts/thread_references.py:11` | `THREAD_DISCUSSION_THREAD_PREFIX` | AgentRun thread references |
| `brain/app/api/routers/cortex/_discussion.py:32` | `THREAD_DISCUSSION_CONVERSATION_PREFIX` | conversation ids |
| `brain/systems/runs/tool_catalog/handlers/chat.py:11` | `THREAD_DISCUSSION_CONVERSATION_PREFIX` | conversation ids |

## How this surfaced

While fixing #812, the same literal was found duplicated across **three** run-thread call sites (`work_intake.py`, `cortex/runner.py`, `recipes/fast.py`). Those were consolidated into one owner in `brain/contracts/thread_references.py`, and all three now import it.

A code-quality review then flagged the two `_CONVERSATION_PREFIX` definitions as a failed consolidation. They are **not** the same thing, which is why #812 deliberately left them alone:

- They carry a different name and a different documented concern (conversation ids, not run thread references).
- They pre-date #812.
- Merging them requires a judgement nobody has made yet.

## The actual question to answer

**Are a "thread-discussion conversation id" and a "thread-discussion run thread reference" the same concept wearing two names, or two concepts that happen to share a prefix?**

- If they are the **same**: delete both local definitions and import `THREAD_DISCUSSION_THREAD_PREFIX` from `brain/contracts/thread_references.py`. The layering test already allows `brain.app → brain.contracts` and `brain.systems → brain.contracts`, so no boundary work is needed.
- If they are **different**: they should not share a literal by coincidence. Give the shared prefix one owner in `brain/contracts/` and have each concern derive its own constant from it, so a future change to the prefix cannot desynchronize them.

Either way the literal ends up with one owner. Today a change to the prefix silently breaks whichever of the three the author did not grep for.

## Acceptance

1. `git grep -nE '"thread-discussion:"'` returns exactly **one** definition.
2. Every consumer imports it (directly, or via a derived constant).
3. A short note in the code or the PR saying which of the two answers above was chosen, and why.
4. `pytest tests/ -m "not requires_db and not requires_browser and not live_provider"` stays green, including `tests/test_architecture_boundaries.py`.

## Notes

[...]

PR #814, opened 2026-08-13
fix(browser): reject non-UUID inbound thread references before they reach asyncpg

### LINKED ISSUE #812: [runtime][bug] browser.open rejects non-UUID inbound_submission thread_id
## Summary
`browser.open` fails in a headless `inbound_submission` when the target `thread_id` is the inbound routing identifier rather than a UUID.

## Evidence
Opening the canonical Google Sheet produced a SQLAlchemy/asyncpg `DataError`: invalid UUID passed as a query argument. The target thread id was:

`inbound:c3c835df-b24d-407f-a5b3-92e8e9f4607d:88e6143d-2e07-4c79-927c-fd319118dc27`

This is not a UUID; it is the inbound form `inbound:<connection_id>:<event_id>`.

## Expected behavior
Browser operations in headless inbound submissions should resolve or normalize the inbound thread reference before querying UUID-typed thread storage, or otherwise use a safe non-UUID path. They must not pass the composite inbound identifier directly to asyncpg as a UUID query argument.

## Impact
Canonical Google Sheet opening fails for inbound submissions, preventing browser-based processing of the submission.

## Reproduction context
- kind: `inbound_submission`
- connection_id: `c3c835df-b24d-407f-a5b3-92e8e9f4607d`
- event_id: `88e6143d-2e07-4c79-927c-fd319118dc27`
- failing target thread_id: `inbound:c3c835df-b24d-407f-a5b3-92e8e9f4607d:88e6143d-2e07-4c79-927c-fd319118dc27`
- observed error: SQLAlchemy asyncpg `DataError` for invalid UUID query argument
- affected operation: `browser.open` against the canonical Google Sheet

redawear:
> *This was generated by AI during a scheduled R3 run.*

Closes #812.

## What was broken

A headless `inbound_submission` run carries a **synthetic, non-UUID** `thread_id`:

```
inbound:c3c835df-b24d-407f-a5b3-92e8e9f4607d:88e6143d-2e07-4c79-927c-fd319118dc27
```

That value reached Postgres UUID columns and asyncpg raised a `DataError`, so `browser.open` failed and the canonical Google Sheet never opened.

The path, end to end:

1. `brain/systems/inbound/service.py:1346` mints `f"inbound:{connection_id}:{event.id}"`. This is the documented design (`docs/personal-agent-connections-mvp.md:766`), not a bug.
2. `brain/systems/runs/cortex/recording.py:217` copies `run.thread_id` into the agent context's `idea_id`.
3. `brain/platform/browser/service.py` compared that against `Idea.id` and `BrowserSession.idea_id` — both `UUID(as_uuid=False)`. Boom.

## Why it fails clearly instead of resolving

`browser_sessions.idea_id` is **NOT NULL with an FK to `ideas.id`**, and an inbound submission has **no `ideas` row at all**. There is no UUID hiding inside the composite id that points at a real Idea, so normalization cannot recover one. The ticket explicitly allows "or otherwise use a safe non-UUID path".

So the change fails **before** the query with a typed, actionable error instead of a raw asyncpg traceback.

## The change

One shared resolver, `brain/contracts/thread_references.py`, which turns a run thread reference into either a real idea UUID or an explicit reason it has none:

| input | result |
|---|---|
| a bare UUID | `IDEA` + canonical uuid |
| `thread-discussion:<uuid>` | `IDEA` + the underlying uuid |
| `inbound:<uuid>:<uuid>` | `INBOUND` → raises `IdeaBackedThreadRequiredError` |
| anything else | `NOT_IDEA_BACKED` → raises |

It lives in `brain.contracts` because the layering test allows `brain.platform → brain.contracts` but forbids `brain.platform → brain.systems`. The first attempt put it higher and `tests/test_architecture_boundaries.py` caught it.

It also folded a pre-existing duplication: `THREAD_DISCUSSION_THREAD_PREFIX` was defined in **three** files (`work_intake.py`, `cortex/runner.py`, `recipes/fast.py`) and now has one owner that all three import.

## Verification

The repo's own backend CI lane, run **serially on an unloaded machine** by the orchestrator (not by the worker), twice at this commit:

```
4944 passed, 11 skipped, 90 deselected — 0 failed, 0 errors
```

The diff touches `brain/**` and `tests/**` only, so the frontend lane does not apply.

The implementing worker reported `1 failed, 5 errors`. All six were `PermissionError: [Errno 1]` on `socket.bind` — the Codex sandbox blocking localhost binds. `4938 + 6 = 4944` accounts for them exactly, and they vanish outside the sandbox.

## How to check this without reading the diff

There is no UI here — it is a runtime crash on a headless path. The honest surfaces:

1. **The regression test names the bug.** `tests/test_browser_sessions.py::test_browser_open_rejects_headless_inbound_thread_before_session_query` uses the ticket's **exact** connection and event ids, and asserts `create_session.assert_not_awaited()` — proving no query is issued at all, which is the actual fix.
2. **On staging**, trigger an inbound submission that calls `browser.open`. Before: an asyncpg `DataError` in the run log. After: a one-line error saying the run is a headless inbound submission and browser tools need an idea-backed thread.
3. Normal cortex runs are unchanged — a real UUID `idea_id` takes the same path as before, with no extra query.

## Review findings NOT folded in — please see these

A cross-family Codex code-quality review returned **BLOCKING**. I verified each finding rather than acting on it, and deliberately did not fold two of them:

1. **"The prefix consolidation is incomplete."** *Verified misleading.* The 3→1 consolidation is real and complete. The reviewer found two survivors named `THREAD_DISCUSSION_CONVERSATION_PREFIX` (`chat.py:11`, `_discussion.py:32`) — a different constant for a different concern, pre-dating this ticket. Filed separately as #813 rather than merged blind.
2. **"The guard is scattered across 8 sites."** *Partly fair, not folded.* A fix pass attempted it and stopped on a real contradiction: the reviewer wanted `get_idea_org_id_async` restored to best-effort `None`, but `create_or_get_session` calls it — returning `None` there would let an invalid `idea_id` through to the insert and reproduce this very bug. **Raising is the correct behavior**, so the current shape stands. What remains genuinely redundant is cosmetic: `create_or_get_session` re-resolves an already-resolved value, and two sites validate `record.idea_id` read back out of a UUID column, where it cannot be invalid.
3. **`IdeaThreadReference` permits invalid combinations** — e.g. `(IDEA, None)` is constructible though the resolver never produces it. A `__post_init__` assert would close it. Not folded; low value, and the fix pass was spent on finding 2.

The reviewer's own summary of the change: *"puts thread-reference normalization in an architecture-safe shared layer and fails before an invalid database query."*

### DIFF
--- a/brain/contracts/thread_references.py
+++ b/brain/contracts/thread_references.py
@@ -1,0 +1,76 @@
+"""Shared normalization for AgentRun thread references."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from enum import Enum
+from typing import Any
+from uuid import UUID
+
+
+THREAD_DISCUSSION_THREAD_PREFIX = "thread-discussion:"
+INBOUND_THREAD_PREFIX = "inbound:"
+
+
+class IdeaThreadReferenceKind(str, Enum):
+    """How an AgentRun thread reference relates to an Idea."""
+
+    IDEA = "idea"
+    INBOUND = "inbound"
+    NOT_IDEA_BACKED = "not_idea_backed"
+
+
+class IdeaBackedThreadRequiredError(ValueError):
+    """Raised when an operation needs an Idea-backed AgentRun thread."""
+
+
+@dataclass(frozen=True)
+class IdeaThreadReference:
+    """The normalized Idea identity, or why the thread has none."""
+
+    kind: IdeaThreadReferenceKind
+    idea_id: str | None = None
+
+    def require_idea_id(self, *, operation: str) -> str:
+        if self.idea_id is not None:
+            return self.idea_id
+        if self.kind is IdeaThreadReferenceKind.INBOUND:
+            raise IdeaBackedThreadRequiredError(
+                f"{operation} need an idea-backed thread; this run is a headless inbound submission"
+            )
+        raise IdeaBackedThreadRequiredError(
+            f"{operation} need an idea-backed thread with a valid UUID idea_id"
+        )
+
+
+def resolve_idea_thread_reference(value: Any) -> IdeaThreadReference:
+    """Resolve a run thread reference without sending invalid UUIDs to storage."""
+
+    text = str(value or "").strip()
+    if text.startswith(THREAD_DISCUSSION_THREAD_PREFIX):
+        text = text.removeprefix(THREAD_DISCUSSION_THREAD_PREFIX).strip()
+    elif text.startswith(INBOUND_THREAD_PREFIX):
+        parts = text.split(":")
+        if len(parts) == 3:
+            try:
+                UUID(parts[1])
+                UUID(parts[2])
+            except (TypeError, ValueError):
+                pass
+            else:
+                return IdeaThreadReference(IdeaThreadReferenceKind.INBOUND)
+
+    try:
+        idea_id = str(UUID(text))
+    except (TypeError, ValueError, AttributeError):
+        return IdeaThreadReference(IdeaThreadReferenceKind.NOT_IDEA_BACKED)
+    return IdeaThreadReference(IdeaThreadReferenceKind.IDEA, idea_id)
+
+
+__all__ = [
+    "IdeaBackedThreadRequiredError",
+    "IdeaThreadReference",
+    "IdeaThreadReferenceKind",
+    "THREAD_DISCUSSION_THREAD_PREFIX",
+    "resolve_idea_thread_reference",
+]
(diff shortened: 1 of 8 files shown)


redawear:
**#815 is stacked on this branch.** It resolves #813, the naming follow-up this PR's review surfaced (`"thread-discussion:"` still had three owners under two names).

Merge order: **this PR first**, then #815 — or merge the train in one click. #815 is a pure rename with no behavior change and the full backend suite green at 4944 passed.

PR #832, opened 2026-08-17
test(deploy): bound the worker-swap simulator so the fast suite cannot hang (#827)

### LINKED ISSUE #827: [tests] Fast suite can hang 24h instead of failing — _run() has no timeout and one test sets an 86400s deadline

## Problem

The fast suite can **hang for up to 24 hours instead of failing**. It happened today and cost a full R3 routine session: a `pytest` run started at 09:32:34Z was still wedged at 10:43Z (69 minutes), holding its session blocked and its ticket claim open.

The hang is in `tests/test_worker_swap_deploy.py`. Three facts combine:

1. `_run()` shells out with **no timeout**:
   ```python
   def _run(script: str) -> subprocess.CompletedProcess:
       return subprocess.run(["bash", "-c", script], capture_output=True, text=True)
   ```
2. `test_the_drain_wait_ends_when_the_handoff_worker_dies` deliberately sets `drain_timeout_seconds=86400`, with the comment *"A deadline far beyond any tick this test performs: the wait must end on the lost cover, never because the clock happened to run out."*
3. There is **no global test timeout** — no `timeout` setting in `pytest.ini`/`pyproject.toml`, and `pytest-timeout` is not in `requirements.txt`.

So the test that exists to prove *"the wait ends on the lost cover, not on the clock"* will, whenever that property does not hold, **wait on the clock — for the full 24 hours.** The guard has no bounded failure mode.

## Evidence (2026-08-14, same machine, same commit)

| run | tmpdir | result |
|---|---|---|
| R3 session `00348be3`, started 09:32:34Z | `pytest-1181` | **wedged 69+ min**, 0.0% CPU, 47s CPU total |
| R3 session `401fc876`, started 10:39Z | `pytest-1182` | **5015 passed, 11 skipped, 90 deselected in 119.46s** |

Same command, same worktree, same commit — so this is intermittent, not deterministic. Process state of the wedged run:

```
18319  01:11:25  0:47.49  0.0%  S   (pytest)
20509  01:09:43            1.2%      (bash -c, the simulator script)
```

The wedged child is the simulator shell from `_simulator(...)`, whose `sleep` is a shell-function shim that ticks a counter. When the loop under test does not take the shimmed path, the real deadline (86400s) governs.

This is **not** #657 (a wall-clock assertion that *fails* under load) and not #755 (a maintainability split of `wait_for_worker_exit`). Both are closed and neither bounds the runtime of the suite.

## Why it matters beyond one test

A hang is strictly worse than a failure here. A failing test reports; a hung test is indistinguishable from a slow one, so it silently consumes the session that started it. Any future test in this file that simulates a non-terminating loop inherits the same 24-hour ceiling.

## Acceptance

- `_run()` passes an explicit `timeout=` (a few seconds is ample — the passing run does 24 of these tests inside a 119s whole-suite budget) and a `subprocess.TimeoutExpired` **fails** the test with the captured output.
- The fast suite has a global upper bound, so no single test can wedge it indefinitely (e.g. add `pytest-timeout` to `requirements.txt` and set a per-test timeout in `pytest.ini`).
- A deliberately non-terminating simulator loop produces a **red test in seconds**, not a hung process. Prove it by forcing the loop to never clear `handoff-1.running` and showing the test fails fast.
- `pytest tests/ -m "not requires_db and not requires_browser and not live_provider"` stays green.

redawear:
Closes #827

The fast suite could hang for **24 hours** instead of failing. This bounds it.

## What changed

1. `tests/test_worker_swap_deploy.py` — `_run()` now passes `timeout=5` and, on `subprocess.TimeoutExpired`, fails the test with the **captured stdout and stderr** instead of hanging or raising a bare error.
2. `requirements.txt` — adds `pytest-timeout>=2.4`.
3. `pytest.ini` — `timeout = 120`, a global per-test ceiling so no single test can ever wedge the suite again.
4. A new test proves the fast-fail path reports the captured output.

## Why it mattered more than the ticket said

The ticket named one test. **Six** call sites in that file set `drain_timeout_seconds=86400` — lines 368, 394, 482, 526, 615, 633 — so one suite run could lose up to six days. The deadline in `deploy/scripts/compose-runtime-lib.sh:512-530` uses bash `$SECONDS`, which is wall clock, so the test's no-op `sleep` shim cannot bound it. An explicit `timeout=` is the only thing that works.

This was reproduced live twice while gating an unrelated PR on 2026-08-17, on an idle host: two consecutive tests wedged at 8m50s and 10m22s, both at ~2.5% CPU. A wedged run **looks idle, not busy** — that is why it went unnoticed for 13 hours on 2026-08-14.

## Verification

| gate | before | after |
| --- | --- | --- |
| `tests/test_worker_swap_deploy.py` alone | hangs — up to 6×24h | **25 passed in 2.94s** |
| whole fast suite | never completed | **4957 passed, 11 skipped, 0 failed, 90.31s** |

Forced-hang demonstration: the simulator loop was edited to never clear `handoff-1.running`; the test failed in **5.37s** with the captured output, instead of hanging. That edit is not in the diff.

The whole-suite run above is the first time this suite has completed on this host. Load average was ~6 and no other work was running.

## The trigger is a broken `python3`, and this PR is the guard rather than the cure

Found after the fix landed, and it changes how to read this ticket.

On a **clean `origin/main`** with `/opt/homebrew/bin` ahead of `/usr/bin` on `PATH`, the supposedly-hanging file passes in **2.18 seconds**:

```
git worktree add --detach /tmp/base origin/main
PATH="/opt/homebrew/bin:$PATH" venv/bin/python -m pytest tests/test_worker_swap_deploy.py -q
->  24 passed in 2.18s
```

With the default `PATH` on this host, `/usr/bin` wins and `python3` is **Xcode's 3.9.6**. The simulator shells out to `python3`, which then dies importing `brain/contracts/worker_lifecycle.py`:

```
ImportError: cannot import name 'StrEnum' from 'enum'   # StrEnum is 3.11+
```

The drain loop therefore never observes the worker exiting and spins against the 86400-second wall-clock deadline. That is the 24-hour hang.

So the honest framing: **the hang is an environment fault, and this PR is defense in depth.** It is still worth merging — an unbounded wait is a latent trap whatever causes it, and a bounded red in 5 seconds is strictly better than a 24-hour wedge. But merging it will not, by itself, stop a machine with a bad `PATH` from producing 12 red tests in this file instead of 12 hung ones. The interpreter is the thing to fix, and it belongs in its own ticket.

## Two notes for the reviewer

- **`pytest-timeout` must be installed for the `pytest.ini` ceiling to do anything.** Without the plugin, pytest ignores the unknown `timeout` key with only a warning — a silent no-op. It is in `requirements.txt` now; anyone with an existing virtualenv needs to reinstall.
- The ticket flagged an unexplained observation: the wedged run carried `HANDOFF_DIES_AFTER_TICKS=1`, so the `definitively_not_claiming` fast-exit should have fired and did not. That was investigated and **not** reproduced. It stays unexplained. It does not affect this fix, which converts any such hang into a red test in seconds — but it means the underlying fast-exit path may still have a bug worth its own ticket.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -5,6 +5,7 @@ python_classes = Test*
 python_functions = test_*
 addopts = -v --tb=short
 asyncio_mode = auto
+timeout = 120
 markers =
     integration: marks tests requiring a live database (deselect with '-m "not integration"')
     requires_db: marks tests needing Docker test DB (deselect with '-m "not requires_db"')
--- a/requirements.txt
+++ b/requirements.txt
@@ -40,3 +40,6 @@ accelerate>=1.0
 # production image too; the "Local" voice provider degrades to unavailable if
 # this is somehow missing.
 faster-whisper>=1.0
+
+# Test safety
+pytest-timeout>=2.4
--- a/tests/test_worker_swap_deploy.py
+++ b/tests/test_worker_swap_deploy.py
@@ -12,8 +12,11 @@ from __future__ import annotations
 import subprocess
 from pathlib import Path
 
+import pytest
+
 ROOT = Path(__file__).resolve().parents[1]
 RUNTIME_LIB = ROOT / "deploy" / "scripts" / "compose-runtime-lib.sh"
+SIMULATOR_TIMEOUT_SECONDS = 5
 
 
 def _simulator(
@@ -187,8 +190,48 @@ echo "RUNNING=$(compose ps --status running -q worker 2>/dev/null | sort | paste
 '''
 
 
-def _run(script: str) -> subprocess.CompletedProcess:
-    return subprocess.run(["bash", "-c", script], capture_output=True, text=True)
+def _run(script: str) -> subprocess.CompletedProcess[str]:
+    try:
+        return subprocess.run(
+            ["bash", "-c", script],
+            capture_output=True,
+            text=True,
+            timeout=SIMULATOR_TIMEOUT_SECONDS,
+        )
+    except subprocess.TimeoutExpired as exc:
+        stdout = (
+            exc.stdout.decode(errors="replace") if isinstance(exc.stdout, bytes) else exc.stdout
+        )
+        stderr = (
+            exc.stderr.decode(errors="replace") if isinstance(exc.stderr, bytes) else exc.stderr
+        )
+        pytest.fail(
+            f"worker-swap simulator did not exit within {SIMULATOR_TIMEOUT_SECONDS}s\n"
+            f"--- captured stdout ---\n{stdout or ''}\n"
+            f"--- captured stderr ---\n{stderr or ''}",
+            pytrace=False,
+        )
+
+
+def test_run_fails_with_captured_output_when_simulator_times_out(monkeypatch):
+    def time_out(*args, **kwargs):
+        assert kwargs["timeout"] == SIMULATOR_TIMEOUT_SECONDS
+        raise subprocess.TimeoutExpired(
+            cmd=args[0],
+            timeout=kwargs["timeout"],
+            output=b"simulator stdout\n",
+            stderr=b"simulator stderr\n",
+        )
+
+    monkeypatch.setattr(subprocess, "run", time_out)
+
+    with pytest.raises(pytest.fail.Exception) as failure:
+        _run("exit 0")
+
+    message = str(failure.value)
+    assert "did not exit within 5s" in message
+    assert "simulator stdout" in message
+    assert "simulator stderr" in message
 
 
 def _calls(state: Path) -> str:


redawear:
**Superseded by [PR #836](https://github.com/Illospace/illospace/pull/836) — merge that one instead.**

This work is folded unchanged into the consolidated branch `illo-qa/consolidated-20260817`, together with the other three open R3 PRs. All four merged with zero conflicts and the consolidated tree was gated fresh: **5021 passed, 11 skipped, 0 failed** on the repo's own CI command, one `alembic` head.

Four open PRs made you the conflict resolver three times. #836 is one click. This PR stays open until #836 merges, then it is closed as shipped.

PR #835, opened 2026-08-17
fix(deploy): give the host interpreter pin a deploy-wide owner and an error contract (#834)

### LINKED ISSUE #834: [deploy] Host-side deploy scripts bypass the WORKER_SWAP_PYTHON_BIN pin and fail silently on a stale interpreter

## Problem

`ops/deploy.sh` deliberately pins the host interpreter:

```sh
PYTHON_BIN="$ROOT/venv/bin/python3"
WORKER_SWAP_PYTHON_BIN="$PYTHON_BIN"
```

`deploy/scripts/worker-swap-lib.sh:6` honours that pin. `deploy/scripts/compose-runtime-lib.sh` declares the same variable at line 8 (`WORKER_SWAP_PYTHON_BIN="${WORKER_SWAP_PYTHON_BIN:-python3}"`) and then **ignores it** in `compose_env_value()`, which hardcodes `python3` at line 22. Both files are sourced by the same deploy path, so one half of the subsystem is pinned and the other half is not.

`compose_env_value()` reads `deploy/compose/.env`. It has no error check: if the interpreter is missing or too old, the heredoc dies on stderr and the function prints nothing. The caller cannot tell "the key is absent" from "the interpreter failed". `compose_profile_enabled()` consumes it directly, so a bad interpreter silently reports every compose profile as disabled.

The same hardcoded `python3` appears in four more host-side call sites:

| file | line | context |
|---|---|---|
| `deploy/scripts/compose-runtime-lib.sh` | 22 | `compose_env_value()` |
| `deploy/scripts/doctor.sh` | 131 | public-URL check |
| `deploy/scripts/doctor.sh` | 151 | vault-key check |
| `deploy/scripts/init-secrets.sh` | 77 | env-file rewrite |
| `deploy/scripts/smoke-test.sh` | 63, 91 | smoke assertions |

`doctor.sh:101` (`need_command python3`) checks presence only, never version.

Container-side calls are **correct as they stand** and must not change: `deploy/scripts/doctor.sh:282` and `deploy/scripts/upgrade.sh:145` both run through `compose exec -T`, where `python3` is the image's interpreter. The `deploy/docker/*.Dockerfile` lines are likewise in scope of the image, not the host.

## Why it matters

The failure is silent and gives a wrong answer instead of an error. A host whose `python3` predates the repo's floor (`brain/contracts/worker_lifecycle.py` imports `enum.StrEnum`, so 3.11+) will run the deploy scripts to completion with empty env values.

Observed on the R3 automation host: a stale `python3` on `PATH` made a worker-swap simulation spin against the 86400-second drain deadline in `wait_for_worker_exit()` instead of failing. Issue #827 bounds that wait — a guard on the symptom. This ticket removes the cause.

## Change

1. Route every **host-side** `python3` invocation in `deploy/scripts/` through the existing `WORKER_SWAP_PYTHON_BIN` pin, or through one shared helper that resolves it. Do not introduce a second, competing variable.
2. Make `compose_env_value()` distinguish an interpreter failure from a missing key: check the exit status and fail loudly rather than returning an empty string.
3. Leave every `compose exec -T … python3` call and every Dockerfile line untouched.

## Acceptance

- `git grep -nE '(^|[^_[:alnum:]])python3 ' deploy/scripts/` returns only the two container-side lines (`doctor.sh:282`, `upgrade.sh:145`).
- With `WORKER_SWAP_PYTHON_BIN` pointed at a deliberately broken interpreter, `compose_env_value` exits non-zero and prints a diagnostic naming the binary — it does not return an empty value.
- `compose_profile_enabled` keeps its current behaviour when the interpreter is healthy and the key is genuinely absent.
- A regression test covers the interpreter-failure path (the existing deploy tests in `tests/test_worker_swap_deploy.py` and `tests/test_safe_deploy.py` are the pattern to follow).

redawear:
Closes #834

## What this changes

`ops/deploy.sh` pinned the host interpreter, but only the worker-swap layer honoured the pin — the other host-side deploy scripts called a bare `python3`. Worse, `compose_env_value()` swallowed an interpreter failure and returned an empty string, which `compose_profile_enabled` read as **every compose profile disabled**. A wrong answer, silently, on a deploy path.

Two commits:

**1. Route every host-side call through the pin** (`5426630f`) — `compose-runtime-lib.sh`, `doctor.sh`, `init-secrets.sh`, `smoke-test.sh`. `compose_env_value()` now exits non-zero with a stderr line naming the binary.

**2. Give the contract an owner and an explicit error code** (`bf4a4501`) — folds in the maintainability review below.

- New `deploy/scripts/deploy-python-lib.sh` owns `DEPLOY_PYTHON_BIN` and `deploy_python_bin()`. The worker-swap lib now *consumes* that contract instead of owning it — `init-secrets.sh` and `smoke-test.sh` do no worker swapping and should never have had to source it.
- **Nothing breaks.** `DEPLOY_PYTHON_BIN="${DEPLOY_PYTHON_BIN:-${WORKER_SWAP_PYTHON_BIN:-python3}}"`, and `worker_swap_python_bin()` stays as a delegating wrapper, so any caller still setting the old name keeps working.
- `compose_profile_enabled` / `compose_service_enabled` now carry a stated contract: **`0` enabled, `1` disabled, `>=2` operational error**. `upgrade.sh` checks it and aborts with a message instead of quietly deciding meetbot is off.
- The interpreter is resolved once per script rather than by a command substitution at each call site.

**Container-side calls are deliberately untouched** — `doctor.sh:287` and `upgrade.sh:152` run through `compose exec -T`, where `python3` is the image's interpreter. The Dockerfiles are in the image's scope too.

## Why it mattered

Found the hard way. On the R3 automation host a stale `python3` made a worker-swap simulation spin against the 86400-second drain deadline in `wait_for_worker_exit()` instead of failing. #827 bounds that wait — a guard on the symptom. This removes the cause.

## How to check it — no UI, two commands

```
git grep -nE '(^|[^_[:alnum:]])python3 ' -- deploy/scripts/
```
Expect exactly two lines, both container-side: `doctor.sh:287`, `upgrade.sh:152`.

```
DEPLOY_PYTHON_BIN=/nonexistent/python bash -c \
  'source deploy/scripts/compose-runtime-lib.sh; compose_env_value COMPOSE_PROFILES; echo "exit=$?"'
```
Expect `exit=2` and a stderr line naming `/nonexistent/python`. Before this PR the same command printed an empty line and exited 0.

Then confirm the old name still works — swap `DEPLOY_PYTHON_BIN` for `WORKER_SWAP_PYTHON_BIN` in that command and expect the identical result.

## Verification

- **Full fast suite on the final tree: 4958 passed, 11 skipped, 0 failed, 112s.** Run serially at load 3.7 with zero other Codex workers on the machine, so the number is not a contention artifact.
- `bash -n` clean on every changed shell file, including `illo` and `ops/deploy.sh`.
- Tests cover both directions: the interpreter-failure path fails loudly through the **public** `compose_service_enabled` decision, and a healthy interpreter with a genuinely absent key still returns the quiet "disabled" answer.

## Review

A cross-family maintainability review (Codex) returned **BLOCKING** on the first commit with five findings. Four are folded into `bf4a4501`: the ownership/naming of the contract, the error-vs-disabled conflation, the repeated command substitutions, and the missing public-path test coverage.

The fifth is **declared out of scope**: splitting the 1,334-line `tests/test_safe_deploy.py` into a focused module. That module's size predates this ticket, and moving tests would bury this change's diff. Worth its own ticket if it keeps growing.

### One residual note

`deploy-python-lib.sh` resolves `DEPLOY_PYTHON_BIN` at **source time**. Every caller in this repo sets it before sourcing, so nothing here is affected — but an external caller that sets `WORKER_SWAP_PYTHON_BIN` *after* sourcing would no longer be honoured, where the old call-time lookup would have caught it. Flagging rather than guessing at an external contract.

### DIFF
--- a/deploy/scripts/compose-runtime-lib.sh
+++ b/deploy/scripts/compose-runtime-lib.sh
@@ -5,8 +5,8 @@ ROOT="${ROOT:-$(cd "$COMPOSE_RUNTIME_LIB_DIR/../.." && pwd)}"
 COMPOSE_FILE="${COMPOSE_FILE:-$ROOT/deploy/compose/docker-compose.yml}"
 ENV_FILE="${ENV_FILE:-${ILLO_COMPOSE_ENV_FILE:-$ROOT/deploy/compose/.env}}"
 RUNTIME_SERVICE_CATALOG="${RUNTIME_SERVICE_CATALOG:-$ROOT/deploy/compose/runtime-services.json}"
-WORKER_SWAP_PYTHON_BIN="${WORKER_SWAP_PYTHON_BIN:-python3}"
 
+source "$COMPOSE_RUNTIME_LIB_DIR/deploy-python-lib.sh"
 source "$COMPOSE_RUNTIME_LIB_DIR/worker-swap-lib.sh"
 source "$COMPOSE_RUNTIME_LIB_DIR/worker-lifecycle-lib.sh"
 
@@ -19,7 +19,7 @@ compose() {
 
 compose_env_value() {
   local key="$1" python_bin
-  python_bin="$(worker_swap_python_bin)"
+  python_bin="$(deploy_python_bin)"
   if ! "$python_bin" - "$ENV_FILE" "$key" <<'PY'
 from pathlib import Path
 import re
@@ -42,10 +42,12 @@ for line in path.read_text(encoding="utf-8").splitlines():
 PY
   then
     echo "Failed to read $key from $ENV_FILE with Python interpreter: $python_bin" >&2
-    return 1
+    return 2
   fi
 }
 
+# Exit status contract for compose_profile_enabled and compose_service_enabled:
+# 0 means enabled, 1 means disabled, and 2 or greater means an operational error.
 compose_profile_enabled() {
   local profile="$1" profiles
   if [ "${COMPOSE_PROFILES+x}" = "x" ]; then
@@ -62,8 +64,13 @@ compose_profile_enabled() {
 }
 
 compose_service_enabled() {
-  local service="$1" profile="${2:-$1}"
-  compose_profile_enabled "$profile" && return 0
+  local service="$1" profile="${2:-$1}" profile_status
+  if compose_profile_enabled "$profile"; then
+    return 0
+  else
+    profile_status=$?
+  fi
+  [ "$profile_status" -eq 1 ] || return "$profile_status"
   [ -n "$(compose_service_container_id "$service")" ]
 }
 
--- a/deploy/scripts/compose-runtime-lib.sh
+++ b/deploy/scripts/compose-runtime-lib.sh
@@ -18,8 +18,9 @@ compose() {
 }
 
 compose_env_value() {
-  local key="$1"
-  python3 - "$ENV_FILE" "$key" <<'PY'
+  local key="$1" python_bin
+  python_bin="$(worker_swap_python_bin)"
+  if ! "$python_bin" - "$ENV_FILE" "$key" <<'PY'
 from pathlib import Path
 import re
 import sys
@@ -39,6 +40,10 @@ for line in path.read_text(encoding="utf-8").splitlines():
     print(value)
     break
 PY
+  then
+    echo "Failed to read $key from $ENV_FILE with Python interpreter: $python_bin" >&2
+    return 1
+  fi
 }
 
 compose_profile_enabled() {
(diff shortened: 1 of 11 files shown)


redawear:
**Superseded by [PR #836](https://github.com/Illospace/illospace/pull/836) — merge that one instead.**

This work is folded unchanged into the consolidated branch `illo-qa/consolidated-20260817`, together with the other three open R3 PRs. All four merged with zero conflicts and the consolidated tree was gated fresh: **5021 passed, 11 skipped, 0 failed** on the repo's own CI command, one `alembic` head.

Four open PRs made you the conflict resolver three times. #836 is one click. This PR stays open until #836 merges, then it is closed as shipped.

PR #841, opened 2026-08-18
refactor(cycles): one shared clause for due illo-lane cycles

### LINKED ISSUE #840: [cycles] The "due illo-lane cycle" filter is spelled six times — promote it into one shared clause helper

Surfaced while reviewing [#818](https://github.com/Illospace/illospace/issues/818) / PR #831. Recorded there as a follow-up, not a merge blocker.

## Problem

The "due illo-lane cycle" filter is a five-predicate clause:

```python
Cycle.deleted_at.is_(None),
Cycle.enabled.is_(True),
Cycle.executor_binding == ILLO_LANE_EXECUTOR_BINDING,
Cycle.next_run_at.is_not(None),
Cycle.next_run_at <op> <cutoff>,
```

It is written out **six times** on `main`, across three modules:

| site | purpose | cutoff |
|---|---|---|
| `brain/app/ops/runtime_status.py:258` | `overdue_cycle_count` | `< captured_at` |
| `brain/app/ops/runtime_status.py:269` | `oldest_overdue_cycle_at` | `< captured_at` |
| `brain/app/ops/runtime_status.py:281` | `overdue_cycles` sample | `< captured_at` |
| `brain/systems/cycles/health.py:72` | `due_cycle_clause` (already factored **locally**) | `<= stale_cutoff` |
| `brain/systems/cycles/service.py:663` | claims due cycles, `with_for_update(skip_locked=True)` | `<= now` |
| `brain/systems/cycles/service.py:751` | catch-up sweep | `< catch_up_before` |

The only thing that varies between copies is the **comparison operator** and the **cutoff value**. Everything else is identical.

`brain/systems/cycles/health.py:72` already factors its own copy into a local `due_cycle_clause` and reuses it twice in the same function — the pattern is proven, it is just not shared.

A seventh, related site is a strict subset (no `next_run_at` predicate) and is out of scope for this ticket: `brain/systems/cycles/service.py:824` (wake-by-name), which filters enabled + not-deleted + illo-lane only.

## Why it matters

`brain/systems/cycles/common.py` is already the owner of this vocabulary — it holds `ILLO_LANE_EXECUTOR_BINDING`, `PERSONAL_AGENT_EXECUTOR_BINDING`, `VALID_CYCLE_EXECUTOR_BINDINGS`, `validate_executor_binding()` and `cycle_executor_binding()`. The constants are shared; the predicate that uses them is not. Every new reader of "which cycles are due" copies five lines and can silently get one wrong.

**A boundary inconsistency already exists and this refactor makes it visible.** The ops surface counts a cycle overdue with `<`, while the scheduler claims it with `<=`. At the exact instant `next_run_at == cutoff`, `runtime_status.py` reports zero overdue cycles while `service.py` considers that cycle claimable. This ticket does **not** ask for a behaviour change — keep each call site's current operator. Just make the difference explicit at the call site instead of buried in a repeated block, so a later ticket can decide it deliberately.

## Change

1. Add one clause helper to `brain/systems/cycles/common.py` (the established home), parameterised by cutoff and by inclusive/strict comparison. It returns a SQLAlchemy `and_(...)`, so every caller keeps composing it into its own `select(...)`.
2. Replace all six copies with a call to it, preserving each site's **current** operator exactly.
3. Delete the now-redundant local `due_cycle_clause` in `health.py` in favour of the shared one.

Keep it a clause helper. Do not introduce a policy/strategy object: seven of these are SQL `WHERE` predicates and the database must still filter the rows.

## Acceptance

- `brain/systems/cycles/common.py` exports one due-cycle clause helper, and no module writes the five predicates inline any more.
- Each of the six call sites keeps the operator it has today (`<` at the three `runtime_status.py` sites and `service.py:751`; `<=` at `health.py` and `service.py:663`).
- `service.py:663` keeps `with_for_update(skip_locked=True)` and its ordering/limit — the claim path's locking behaviour is unchanged.
- The repo's own suite stays green: `pytest tests/ -m "not requires_db and not requires_browser and not live_provider"`.
- A test asserts the strict/inclusive boundary: a cycle whose `next_run_at` equals the cutoff is excluded by the strict form and included by the inclusive form.

redawear:
Closes #840

## What this does

The "due illo-lane cycle" filter was five predicates written out **six times** across three modules. This replaces all six with one shared clause helper. No behaviour changes.

Each call site keeps the operator it has today. The helper takes an explicit `inclusive` flag, so the boundary is now visible at the call site instead of buried in a repeated block:

| site | operator | call |
|---|---|---|
| `runtime_status.py` (3 sites) | `<` | `inclusive=False` |
| `health.py` | `<=` | `inclusive=True` |
| `service.py` claim path | `<=` | `inclusive=True` |
| `service.py` catch-up sweep | `<` | `inclusive=False` |

The pre-existing boundary inconsistency the ticket describes (ops counts overdue with `<`, the scheduler claims with `<=`) is **preserved exactly**, not resolved. It is now one readable argument, so a later ticket can decide it deliberately.

## One deviation from the ticket, on purpose

The ticket asks for the helper in `brain/systems/cycles/common.py`. It went into a new `brain/systems/cycles/queries.py` instead.

A Codex quality review blocked the `common.py` placement: that module holds constants and pure value helpers, and **20 modules import it** — including `app/api/schemas/cycles.py` and `behavior_policy_contract.py`. Adding SQLAlchemy and `Cycle` imports there would make every pure schema and contract consumer inherit an ORM dependency. `common.py` now imports no SQLAlchemy at all.

The ticket's intent — one shared helper, no inline copies — is met. Only its named file changed.

## Acceptance, verified

- **No module writes the five predicates inline.** The only remaining `executor_binding == ILLO_LANE_EXECUTOR_BINDING` sites are the helper itself and `service.py:818`, the wake-by-name filter the ticket explicitly scopes out (it has no `next_run_at` predicate).
- **Every operator preserved** — table above.
- **Claim path locking unchanged**: `with_for_update(skip_locked=True)`, ordering and limit all intact at `service.py:667`.
- **Boundary test added and passing**: a cycle whose `next_run_at` equals the cutoff is excluded by the strict form and included by the inclusive form.
- **Backend suite green**: `5036 passed, 11 skipped, 90 deselected in 107.39s`, run serially at load 9.3 with no other suite running. The diff selects the backend lane only (`brain/**`, `tests/**`); the frontend lane is not triggered. The CI pre-step `check_cycle_context_admission` also passes.

## What to eyeball

This is a pure refactor with no UI surface. The thing worth a look is the operator table above — it is the one thing this change could have broken silently, and it is the reason the boundary test exists.

Two non-blocking notes from the review, left as-is:
- `is_not(None)` in the helper is redundant (either comparison already excludes NULL), kept because it documents scheduled eligibility.
- `service.py:818` correctly stays inline — it selects a *wakeable* cycle, not a *due* one.

### DIFF
--- a/brain/app/ops/runtime_status.py
+++ b/brain/app/ops/runtime_status.py
@@ -14,7 +14,7 @@ from brain.app.scheduler.daemon import async_scheduler_health_snapshot
 from brain.app.scheduler.stale_run_reaper import agent_run_maintenance_snapshot
 from brain.platform.db.models.agent_run import AgentRunRow
 from brain.platform.db.models.cycle import Cycle
-from brain.systems.cycles.common import due_illo_lane_cycle_clause
+from brain.systems.cycles.queries import due_illo_lane_cycle_clause
 from brain.systems.runs.cortex.queue_health import (
     QueueHealth,
     queued_backlog_health_snapshot_async,
--- a/brain/app/ops/runtime_status.py
+++ b/brain/app/ops/runtime_status.py
@@ -14,7 +14,7 @@ from brain.app.scheduler.daemon import async_scheduler_health_snapshot
 from brain.app.scheduler.stale_run_reaper import agent_run_maintenance_snapshot
 from brain.platform.db.models.agent_run import AgentRunRow
 from brain.platform.db.models.cycle import Cycle
-from brain.systems.cycles.common import ILLO_LANE_EXECUTOR_BINDING
+from brain.systems.cycles.common import due_illo_lane_cycle_clause
 from brain.systems.runs.cortex.queue_health import (
     QueueHealth,
     queued_backlog_health_snapshot_async,
@@ -254,36 +254,20 @@ async def async_runtime_status_snapshot(
         await session.scalar(
             select(func.count())
             .select_from(Cycle)
-            .where(
-                Cycle.enabled.is_(True),
-                Cycle.deleted_at.is_(None),
-                Cycle.executor_binding == ILLO_LANE_EXECUTOR_BINDING,
-                Cycle.next_run_at.is_not(None),
-                Cycle.next_run_at < captured_at,
-            )
+            .where(due_illo_lane_cycle_clause(captured_at, inclusive=False))
         )
         or 0
     )
     oldest_overdue_cycle_at = await session.scalar(
         select(func.min(Cycle.next_run_at)).where(
-            Cycle.enabled.is_(True),
-            Cycle.deleted_at.is_(None),
-            Cycle.executor_binding == ILLO_LANE_EXECUTOR_BINDING,
-            Cycle.next_run_at.is_not(None),
-            Cycle.next_run_at < captured_at,
+            due_illo_lane_cycle_clause(captured_at, inclusive=False)
         )
     )
     overdue_cycles = list(
         (
             await session.scalars(
                 select(Cycle)
-                .where(
-                    Cycle.enabled.is_(True),
-                    Cycle.deleted_at.is_(None),
-                    Cycle.executor_binding == ILLO_LANE_EXECUTOR_BINDING,
-                    Cycle.next_run_at.is_not(None),
-                    Cycle.next_run_at < captured_at,
-                )
+                .where(due_illo_lane_cycle_clause(captured_at, inclusive=False))
                 .order_by(Cycle.next_run_at.asc())
                 .limit(20)
             )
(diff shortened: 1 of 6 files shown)


redawear:
Folded into **#853**, which carries this branch plus the four other open slices in one merge click off current `origin/main` (`d35f2685`). Gate on the merged tree: `5007 passed, 0 failed` (the repo's own CI backend lane, run serially at load 5.1), one alembic head, admission check `ok`.

This PR stays open until #853 merges. I will close it then, naming #853.

redawear:
Superseded by #859, which merged to main at 2026-08-28T14:18:27Z (squash commit b2d5b7ff). This PR's content is verified on origin/main by a content probe, so it is closed as shipped, not abandoned.

Issue #855, opened 2026-08-27
Preservation run claims GitHub comment but returns missing durable evidence


axel-havard:
## Summary

A `preserve_knowledge` submission completed and reported that it updated a canonical GitHub issue comment, but the preservation evidence contract remained unsatisfied.

## Event

- Event ID: `182c0cbb-268c-4d1d-b66a-abb218a9e0ec`
- Run ID: `19794`
- Origin: `codex.session-preserve`
- Idempotency key: `media-ingestion-v11-16-2-failure-diagnosis-20260827`

## Exact polling call

```json
{
  "event_id": "182c0cbb-268c-4d1d-b66a-abb218a9e0ec",
  "include_payload": false,
  "limit": 5
}
```

Tool: `illo_get_result`

## Observed

- `status: "review_required"`
- `handling_status: "needs_action"`
- `run_status: "completed"`
- `evidence_status: "missing"`
- `mutated_target_refs: []`
- Attribution says `add_github_issue_comment` ran and the final answer claims the diagnosis was preserved on issue #1884.

## Expected

If the GitHub issue comment counts as durable preservation, the receipt should include a non-empty durable mutation reference and report `evidence_status: "satisfied"`. Otherwise, the run should not claim preservation succeeded and should use one of the accepted durable-storage tools before completion.

PR #853, opened 2026-08-25
consolidated: merged-PR status, due-cycle clause, notify-cycle deletion, deploy prose fields, meetbot silence, specs archival (#838 #840 #843 #845 #847 #846)

redawear:
Six open pull requests folded into one branch off current `origin/main`
(`d35f2685`), so this repo takes **one merge click** instead of six — and so the
duplicate pair on #838 and the conflict on #846 are resolved here rather than by
you at merge time.

Closes #838
Closes #840
Closes #843
Closes #845
Closes #847

## What is in it

| slice | ticket | what it does |
|---|---|---|
| a merged PR must not read as closed | [#838](https://github.com/Illospace/illospace/issues/838) | the webhook envelope gains `pr_outcome` (`merged` / `closed_unmerged` / `open`) and the summary says "merged"; the triage skill stops mapping a merged PR to `Canceled` |
| one shared due-cycle clause | [#840](https://github.com/Illospace/illospace/issues/840) | six copies of the same five-predicate filter collapse into `brain/systems/cycles/queries.py` |
| delete the never-registered notify cycle | [#843](https://github.com/Illospace/illospace/issues/843) | −674 lines: `change_notifications.py`, `change_notifications_cycle.py` and their tests |
| retired deploy fields lost their call site | [#845](https://github.com/Illospace/illospace/issues/845) | drops the dead constant and gives the knowledge connector a real call site, so cached `search_text` stops leaking retired fields |
| meetbot silence guarantee | [#847](https://github.com/Illospace/illospace/issues/847) | corrects an overstated comment and adds a standalone `getUserMedia` probe |
| archive three finished specs | [#846](https://github.com/Illospace/illospace/pull/846) | `specs/meetbot`, `specs/illo-lifecycle` and `specs/illo-handoff-packets` move to `specs/done/` as rationale READMEs; every reference to them is updated |

#846 is a pull request, not an issue, so it has no `Closes` line. It closes itself
when this merges, because its commits become ancestors of `main`.

## The #838 duplicate, and how I resolved it

[#839](https://github.com/Illospace/illospace/pull/839) and
[#851](https://github.com/Illospace/illospace/pull/851) both implement #838. Their
production diffs on `brain/systems/inbound/github_webhook.py` are **byte-identical**
(both blobs are `2f4aa88a`) — only the skill wording and the tests differ. Merging
either one would have left the other conflicting on four files.

This branch takes **#851's** `SKILL.md` and `docs/329-live-delta.md` verbatim, because
[...]

PR #858, opened 2026-08-27
refactor(attribution): emit GitHub refs at the tool boundary (#857)

### LINKED ISSUE #857: [attribution] Move GitHub ref emission to the tool boundary instead of shape-sniffing in the generic ref walker

## The problem

`_add_github_artifact_ref` in `brain/systems/inbound/attribution.py` sniffs GitHub connector payload shapes from inside the **generic** recursive ref walker. It now knows three of them:

- `{"repo", "issue": {"type", "number"}}`
- `{"repo", "pull_request": {"type", "number"}}`
- `{"repo", "issue_number", "comment": {"id"}}`

Every time a GitHub tool grows or changes its return payload, this generic walker has to learn the new shape. It is the only place where the walker hardcodes one connector's contract, and it has grown once per artifact kind so far.

## Why this is the wrong owner

`attribution.py` already has the right seam: `_EXPLICIT_REF_KEYS` (`target_refs`, `mutated_target_refs`, `durable_refs`, `durable_target_refs`). A tool result that carries one of those keys gets its refs collected **without the walker knowing anything about the payload**.

So the durable design is to emit explicit refs at each GitHub tool's own boundary — where the payload is already being constructed and its shape is known — and let the existing explicit-ref collector do the rest. The shape-sniffing in the walker can then be deleted, not extended again.

## Where it came from

Found by a Codex code-quality review of #855 / PR #856, which raised it as BLOCKING on abstraction ownership. It was **not** folded into that PR: #855 is a receipt-correctness bug, and rewiring how every GitHub tool emits refs is a cross-cutting change to the tool handlers. Filing it rather than smuggling it in.

## Acceptance

- GitHub tool handlers (at minimum `add_github_issue_comment`, `create_github_issue`, and the pull-request creator in `brain/systems/runs/tool_catalog/handlers/github.py`) emit explicit refs on their own results.
- The shape-specific branches in `_add_github_artifact_ref` are removed, not merely bypassed.
- Ref **kinds** and **id encodings** stay byte-identical to today's — `github_issue`, `github_pull_request`, `github_issue_comment`, and `owner/repo#N` / `owner/repo#N:comment:ID`. The formatters `_github_artifact_ref_id` / `_github_comment_ref_id` added in #856 are the canonical spelling; move them, do not fork them.
- The existing attribution tests keep passing unchanged. They assert observable refs, so they are the regression net for this refactor.
- Preservation semantics are untouched: `github_issue_comment` must stay OUT of `PRESERVATION_ACCEPTABLE_TARGET_KINDS`.

## Verify

```
venv/bin/python -m pytest -m "not requires_db and not requires_browser and not live_provider" -q
```

No path argument — `pytest.ini` sets `testpaths = tests meetbot/tests`, and naming a path silently drops `meetbot/tests`.

## Not urgent

Nothing is broken today. This is about where the next GitHub payload change lands. Do it before the fourth shape, not after.

redawear:

Closes #857

## What this changes

`_add_github_artifact_ref` sniffed GitHub payload shapes from inside the **generic**
recursive ref walker in `brain/systems/inbound/attribution.py`. It had learned three
shapes and grew one branch per artifact kind — the only place that walker hardcoded a
single connector's contract.

Now each GitHub tool handler emits its own explicit ref, where the payload is built
and its shape is already known. The existing `_EXPLICIT_REF_KEYS` collector does the
rest, and the shape-sniffing is **deleted**, not bypassed.

| Handler | Ref emitted |
|---|---|
| `_handle_create_github_issue` | `github_issue` |
| `_handle_update_github_issue` | `github_issue` |
| `_handle_create_github_pull_request` | `github_pull_request` |
| `_handle_add_github_issue_comment` | `github_issue_comment` |

`mutated_target_refs` is the key in every case, because each of these handlers created
or mutated the artifact it names. Read-only handlers are untouched and still emit
nothing.

`brain/contracts/github.py` owns the encoding: `github_artifact_ref_id`,
`github_comment_ref_id`, and typed factories `github_issue_ref`,
`github_pull_request_ref`, `github_issue_comment_ref`. Kinds and ids are
**byte-identical** to before — `owner/repo#N` and `owner/repo#N:comment:ID`.

## How to check it without reading the diff

Illo run **19794** is the incident this line of work came from: it wrote a GitHub issue
comment and then reported `mutated_target_refs: []`. That gap is what #855 fixed and
what this makes structural. After merge, a `preserve_knowledge` run that files an issue
or posts a comment should show that artifact in its `mutated_target_refs` — the same
place #856's receipt fix made visible.

Two things a reviewer may want to confirm by eye:

1. **A GitHub write still names its artifact.** Trigger any run that files an issue or
   posts a comment, and check the run's target refs name `owner/repo#N`.
2. **A comment still does not count as preserved.** `github_issue_comment` stays out of
   `PRESERVATION_ACCEPTABLE_TARGET_KINDS`, so a run whose only write is a GitHub comment
   must still settle as `missing`. That is the designed answer, unchanged here.

## Verification

- `python -m pytest -m "not requires_db and not requires_browser and not live_provider" -q`
  — the workflow's command **verbatim, no path argument**, so `meetbot/tests` runs:
  **5033 passed, 11 skipped, 90 deselected in 88.12s**. Run serially on an idle machine
  (load 2.93), with no Codex worker competing.
- `python -m brain.jobs.check_cycle_context_admission` — `ok: true`, 3 cycles passed.
- Backend lane only: the diff touches `brain/**` and `tests/**`, **no** `frontend/**`
  and no migration, so `brain-ci.yml`'s `paths-filter` does not select the frontend lane.
- `tests/test_inbound_attribution.py` is **unchanged**. It asserts observable refs, so
  it is the regression net for this refactor rather than something to adjust.

## Intended behaviour change

A raw `{"repo", "issue"}` mapping that is **not** a tool result no longer yields a ref
by itself. Covered by `tests/test_github_attribution_boundary.py`.

Tool events stored **before** this change keep working through
`_legacy_github_result_refs` in `attribution.py`. That path is deliberately frozen: it
runs only for the four tool names that predate the refactor, only when the result
carries no `mutated_target_refs`, and its docstring names its retirement trigger —
delete it once no run events predating `0f3fa6f0` remain readable. A new GitHub result
shape belongs in the handler, never there.

## Review history

A cross-family Codex code-quality review of the first commit returned **BLOCKING** on
abstraction ownership, and `061677cb` folds both findings:

- attribution ran the GitHub adapter over *every* stored result with no cutoff → now a
  named, tool-scoped, frozen legacy function with a stated retirement trigger;
- the emitter re-dispatched on a tool name each handler already knew, and mutated the
  caller's dict via `clear()`/`update()` to control key order → deleted;
  handlers now call typed factories and assign the key normally. Key order was not
  load-bearing.

It also caught that the shared formatters were underscore-private while crossing
package boundaries and appearing in `__all__`; they are public now.

## Known, deliberately not fixed here

`brain/systems/cortex/project_context/github.py:99` — the `canonical_ref` property
returns `f"{self.repo}#{self.number}"`, a second implementation of the same encoding.
It **predates this work** (verified on `origin/main`) and sits on a frozen dataclass in
the promotion path, so folding it in would widen this PR beyond the ticket. Named here
rather than changed.


redawear:

**Superseded by #859 — merge that one instead.** This branch's tip `061677cb` is
merged into it unchanged, so nothing here is lost.

#859 is `origin/main` (`66b85103`) plus #853 and this PR, in that order, both with
**zero conflicts**. The two share zero files (48 vs 8 paths), so merge order could not
change the result. Reda asked for this shape on 2026-08-18: one pull request per repo,
one merge click. #859 closes six issues — #857 plus the five from #853.

Gate on the merged tree: `check_cycle_context_admission` `ok: true`, and the backend
fast suite **5011 passed, 11 skipped, 0 failed** with no path argument, so
`meetbot/tests` ran. One alembic head.

I close this once #859 merges, gated on a content probe of `origin/main`.

redawear:
Superseded by #859, which merged to main at 2026-08-28T14:18:27Z (squash commit b2d5b7ff). This PR's content is verified on origin/main by a content probe, so it is closed as shipped, not abandoned.

PR #862, opened 2026-08-28
fix(project-context): run git clones off the worker event loop (#860)

### LINKED ISSUE #860: Worker event loop freezes during project-context git clones; queue-stall watchdog misreads it and restart-loops the worker

## Summary

The queue-stall watchdog killed the worker twice in 8 minutes on 2026-08-28 (14:06:49Z and 14:13:28Z). The root cause is not the queue. Project-context git clones run as a **blocking `subprocess.run` on the worker's asyncio event loop**. Each clone freezes the entire worker — every in-flight agent run, the claim loop, and all event emission — for the duration of the clone. When a run materializes several repos back to back, the freeze exceeds the watchdog's 45s grace, the watchdog reads the frozen loop as a stalled queue, and the worker exits for restart. The restarted worker reclaims the same runs, re-materializes the same repos, and freezes again. That is the restart loop observed today.

## Evidence (illo-dev, 2026-08-28)

- `docker logs illospace-worker-1`: `agent-run worker queue stalled; exiting for restart` at 14:06:49Z and again at 14:13:28Z (RestartCount=2). The second exit hit `runner drain timed out with 1 in-flight run(s); affected run ids: [19992]`.
- In both stall windows the **entire log stream is silent for 2–3 minutes** (14:04:16→14:06:49 and ~14:11:30→14:13:28) — no httpx lines from any of the ~7 concurrent runs. Run 19987 logged `turn 2` at 14:03:27 and its `turn 3` only appeared at 14:16, after two restarts.
- The first silence begins immediately after a GitHub App installation-token mint for `uwear-ai/uwear-mobile-app` — the materializer's per-repo clone credential path.
- Live confirmation on the recovered worker: a project-context clone ran from ~14:20:45 to ~14:21:30, and the worker's last log line stayed frozen at 14:20:56 until 14:21:36 — logs resumed the moment the clone exited. Sampled at 5s intervals via `docker top` + `docker logs --tail 1`.
- Load amplifier: ~7 concurrent runs (cycle fan-out sweeps), each materializing multiple repos (`uwear-website`, `uwear-backend`, `uwear-mobile-app`, …).

## Mechanism

- `brain/systems/cortex/project_context/materializer.py` — `async def _materialize_resource` calls `_clone_github_repo(...)` directly (no `await`, no executor).
- `_clone_github_repo` uses `run_subprocess_sync` (`brain/platform/async_io.py:325`), which is a plain `subprocess.run`. The same file has the async `run_subprocess` (executor-backed via `run_blocking`) that is NOT used here.
- Clone timeout default is **600s** (`_DEFAULT_GIT_CLONE_TIMEOUT_SECONDS`, materializer.py:49). One slow or hung clone can block the loop for 10 minutes.
- The watchdog (`brain/systems/runs/cortex/queue_health.py::evaluate_queue_health`) counts only *recently active* runs as occupied slots. A frozen loop emits no events, so in-flight runs stop counting, capacity looks free, the queued backlog ages past the threshold, and `QueueStallMonitor` fires after 45s grace (`ILLO_AGENT_RUN_QUEUE_STALL_GRACE_SECONDS`). The watchdog runs on the main thread, so it can still fire while the loop is frozen.
- Restart makes it recur: reclaimed runs re-run workspace materialization, so the same clones repeat on the fresh worker.

## Fix

1. Make the clone non-blocking: in `_materialize_resource`, dispatch `_clone_github_repo` off the loop (`await run_subprocess` / `run_blocking` / `asyncio.to_thread`). Audit the sibling sync git calls in the materializer path (`_git_output`, `_existing_git_checkout`, `shutil.rmtree` of large trees) for the same pattern.
2. Consider: the watchdog conflates "loop frozen" with "queue stalled". After (1) it should no longer misfire, but a loop-liveness heartbeat distinct from queue state would make the diagnosis direct next time.

## Acceptance criteria

- A project-context clone in one run does not delay event emission or turn progress in other in-flight runs (observable: no multi-minute silent gaps in worker logs during clones).
- The queue-stall watchdog does not fire during normal multi-repo materialization under ~7 concurrent runs.
- Existing behavior preserved: clone timeout still enforced; failed clones still fall through the token-candidate loop.

redawear:
## What broke

Project-context git clones ran as a blocking `subprocess.run` **on the worker's asyncio event loop**. Each clone froze the entire worker — every in-flight agent run, the claim loop, all event emission — for as long as the clone took. When one run materialized several repos back to back, the freeze outlasted the queue-stall watchdog's 45s grace. The watchdog read the frozen loop as a stalled queue and exited the worker for restart; the restarted worker reclaimed the same runs, re-cloned the same repos, and froze again.

Observed on illo-dev on 2026-08-28: two kills in 8 minutes (14:06:49Z, 14:13:28Z), the log stream silent for 2–3 minutes in each window, and run 19987's `turn 3` only appearing at 14:16 after two restarts.

## The change

`_materialize_resource` now dispatches both blocking git helpers through `run_blocking` (which is `asyncio.to_thread`):

| line | before | after |
|---|---|---|
| 802 | `_existing_git_checkout(...)` | `await run_blocking(_existing_git_checkout, ...)` |
| 850 | `_clone_github_repo(...)` | `await run_blocking(_clone_github_repo, ...)` |

Offloading the **whole helper** rather than just its `subprocess.run` matters: `_clone_github_repo`'s body is `shutil.rmtree` + `mkdir` + `subprocess.run` + two `_git_output` reads, and `_existing_git_checkout` reaches `check_output_sync` the same way. Moving only the subprocess call would have left the rmtree and the git reads on the loop.

It also lets `_clone_github_repo` stay a plain `def`, so **nothing inside it changed** — the clone timeout, the `finally` block that unlinks the askpass script and pops the token from `env`, the token-candidate fallthrough, and `_sanitize_git_error` all behave exactly as before.

**The watchdog is deliberately untouched.** The ticket lists a loop-liveness heartbeat as a "consider", not as acceptance. It belongs in its own change.

## Code-quality review

A Codex thermo-nuclear review (cross-family — the diff was Codex-written, so this was a second, independent Codex pass at `xhigh`) returned **APPROVE_WITH_NOTES**, `BLOCKING_FINDINGS: none`, `BUGS: none`, 6 files read.

It confirmed the seam: *"the call-site seam is correct because `_materialize_resource` is the only production caller… I would ship this seam"*, and that `await run_blocking(...)` is already the established project spelling (`self_update.py` does the same).

Both of its non-blocking notes are folded in as the second commit:

1. **Blocking boundary made legible.** Both offloaded helpers now say in their docstring that they block and that async callers must use `run_blocking`. The seam was right; nothing at the definition told the next caller which side of it they were on.
2. **Test de-flaked.** The first version asserted `ticks == 3` off three 10 ms sleeps and gave the clone a 1 s release window — which would flake on a loaded box, and this machine sits at load ~19 routinely. It is now a plain two-event handshake with a 30 s bound that exists only so a regression fails instead of hanging the suite. No timing assertion remains.

## How I know the test isn't decorative

The new test blocks the faked clone on a `threading.Event` and asserts that a concurrent coroutine still ticks three times *and* is the thing that releases it.

I verified it has teeth rather than assuming it: I reverted `materializer.py` to `origin/main`, kept the test, and ran it.

```
assert released_by_event_loop == [True]
E   assert [False] == [True]
```

`[False]` is the `wait` timing out — on the pre-fix code the frozen loop never reaches the release. Restored the fix and it passes.

I re-ran that same check **after** the de-flaking rewrite rather than assuming the teeth survived it: still `assert [False] == [True]` with the two call sites reverted, and 0.56 s to pass against the fix.

## Verification

- **Backend fast suite, the CI command verbatim, no path argument** (so `meetbot/tests` actually ran): **5012 passed, 11 skipped, 90 deselected, 0 failed** in 87s. That reconciles — `main` gates at 5011 and this adds exactly one test.
- `python -m brain.jobs.check_cycle_context_admission` → `ok: true`, 3 cycles.
- Lanes: the diff touches `brain/**` and `tests/**` only, so `paths-filter` selects the **backend** lane. No `frontend/**` path, so that lane does not apply.
- **Stated, not hidden:** Spotlight held this machine's load at ~19 during the run. Contention manufactures failures, never passes, so a clean 0-failed still holds.

## What to check on staging

This has no UI. Exercise it on **illo-dev**, where the restart loop was observed:

1. `docker logs -f illospace-worker-1` while a run materializes several repos (a cycle fan-out sweep does it naturally).
2. During a clone, confirm the log keeps producing lines from *other* concurrent runs — the old symptom was a total 2–3 minute silence across every run.
3. Confirm `agent-run worker queue stalled; exiting for restart` does not appear, and `RestartCount` stays put.
4. Confirm clones still succeed and a private repo still resolves through its token candidate — the credential path is the part most worth a second look, even though it is untouched.

## Assumptions

- `run_blocking` is the right seam because `_clone_github_repo` is blocking end to end, not just at its subprocess call.
- The two dirty-worktree helpers elsewhere in the file (`check_output_sync`, `run_subprocess_sync` imports remain) are reached from sync paths; the worker checked and left them alone deliberately.

Closes #860

### DIFF
--- a/brain/systems/cortex/project_context/materializer.py
+++ b/brain/systems/cortex/project_context/materializer.py
@@ -393,7 +393,11 @@ def _git_output(cwd: Path, *args: str) -> str | None:
 
 
 def _existing_git_checkout(destination: Path, slug: str, branch: str | None) -> dict[str, str] | None:
-    """Return metadata for an existing checkout without mutating local state."""
+    """Return metadata for an existing checkout without mutating local state.
+
+    BLOCKING: reaches ``check_output_sync`` via ``_git_output``. Async callers
+    MUST go through ``run_blocking`` (see #860).
+    """
 
     if not destination.exists():
         return None
@@ -444,6 +448,12 @@ def _clone_github_repo(
     token: str | None,
     branch: str | None,
 ) -> dict[str, str]:
+    """Clone a GitHub repo. BLOCKING: rmtree, mkdir, git subprocess, git reads.
+
+    Async callers MUST go through ``run_blocking``. Calling this directly from a
+    coroutine freezes the whole worker for the clone's duration, which trips the
+    queue-stall watchdog and restart-loops the process (see #860).
+    """
     if destination.exists():
         shutil.rmtree(destination, ignore_errors=True)
     destination.parent.mkdir(parents=True, exist_ok=True)
--- a/brain/systems/cortex/project_context/materializer.py
+++ b/brain/systems/cortex/project_context/materializer.py
@@ -12,7 +12,7 @@ import tempfile
 from typing import Any
 
 from brain.contracts.github import parse_github_repo_slug
-from brain.platform.async_io import check_output_sync, run_subprocess_sync
+from brain.platform.async_io import check_output_sync, run_blocking, run_subprocess_sync
 from brain.platform.db.models.run import AgentRun
 from brain.platform.db.repositories.unit_of_work import UnitOfWork
 from brain.systems.cortex.project_context.drafts import load_draft_metadata, sync_draft_from_root
@@ -799,7 +799,7 @@ async def _materialize_resource(
 
     destination = _safe_repo_destination(workspace_root, slug)
     branch = _clean_text(resource.get("branch") or resource.get("default_branch") or resource.get("branch_hint"))
-    existing_checkout = _existing_git_checkout(destination, slug, branch)
+    existing_checkout = await run_blocking(_existing_git_checkout, destination, slug, branch)
     if existing_checkout:
         clone = existing_checkout
         repo_path = Path(clone["path"])
@@ -847,7 +847,7 @@ async def _materialize_resource(
             errors.append(prefix + token_error)
             continue
         try:
-            clone = _clone_github_repo(slug, destination, token=token, branch=branch)
+            clone = await run_blocking(_clone_github_repo, slug, destination, token=token, branch=branch)
         except Exception as exc:
             prefix = f"Vault key {key_name}: " if key_name else "public clone: "
             errors.append(prefix + _sanitize_git_error(str(exc)))
--- a/tests/test_project_context_materializer.py
+++ b/tests/test_project_context_materializer.py
@@ -1,4 +1,6 @@
+import asyncio
 from pathlib import Path
+import threading
 from types import SimpleNamespace
 from unittest.mock import AsyncMock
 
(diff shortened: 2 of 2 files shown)


redawear:
> *This was generated by AI during a routine run.*

Folded into **https://github.com/Illospace/illospace/pull/866** so the three ready fixes are one merge click instead of three.

#866 is cut from current `main` (`d2978080`), carries this PR's commits unchanged, and is gated on the merged tree: **5028 passed, 0 failed** against a 5014 baseline. It also carries a third fix (#865), which shares `tests/test_inbound_reconciliation.py` with PR #864 — folding them here resolves that overlap once, rather than leaving it for the second merge.

Merge #866; this PR can then be closed. Left open for now, and it stays mergeable on its own if you prefer it that way.

redawear:
> *This was generated by AI during a routine run.*

Closing as **shipped**. [#866](https://github.com/Illospace/illospace/pull/866) merged to `main` at 2026-08-31T12:56:12Z (merge commit `33b4811b`) and carries this PRs content.

Verified by content probe on `origin/main`, not by merge status:

- `brain/systems/cortex/project_context/materializer.py:860` — `clone = await run_blocking(_clone_github_repo, ...)`
- `_existing_git_checkout` present, reached through `run_blocking`
- `tests/test_project_context_materializer.py` present on `main`

Control probes ran both ways so an empty result could not read as a clean one. [#860](https://github.com/Illospace/illospace/issues/860) auto-closed on the same merge.

Issue #861, opened 2026-08-28
Preservation run fails with opaque internal error and no tool calls; identical resubmit succeeds


redawear:
> *This was generated by AI during triage.*

## What happened

A `preserve_knowledge` submission failed with an opaque internal error and made no workspace change. An identical resubmission (same payload, new idempotency key) succeeded ~9 minutes later. The content was ordinary (a ~3.5 KB plain-text investigation summary, no parts/attachments).

- **Failed:** event `f4130cfb-758a-461e-99e2-b2404c9d6859`, run `20006`. Submitted 2026-08-28T14:34:27Z via `illo_submit` (`source_tool=claude-code`, `origin=claude-code.session-preserve`, `desired_outcome=preserve_knowledge`, idempotency key `…-preserve-1`). Run went `queued → running → failed` at 14:41:31Z. Failure: `{"category": "internal", "message": "I failed on this and it is still open — I will come back."}`. `evidence_status: failed`, attribution tags `["failed", "no_workspace_change"]`, `mutated_target_refs: []`, `tool_names: []` (the run apparently executed no tools at all).
- **Succeeded:** event `39bde949-182a-49a6-bd28-ef5904b24c70`, run `20013`, same payload, idempotency key `…-preserve-2`. Completed 14:46:02Z, `evidence_status: satisfied`, memory node `2852` / assertion `1106` created.

## Expected

Either the run completes with durable evidence, or the failure receipt says what failed (tool name, error) so the caller can distinguish "retry me" from "payload problem". Two sub-issues:

1. **Opaque failure category.** `category: "internal"` with the first-person message "I failed on this and it is still open — I will come back" carries no diagnostic. The run's attribution shows zero tool calls, so whatever failed happened before/outside tool execution.
2. **"I will come back" is not actionable.** The message implies Illo will self-retry the open item, but the run status is terminally `failed` and `illo_get_result` gives the caller no retry linkage (`retry_attempt`, `replacement_run_id`, `retry_lineage` all null). A caller following the standard verify-receipt discipline must treat it as not-saved and resubmit — which risks a duplicate memory if Illo also "comes back" later.

## Observed client behavior

Caller polled `illo_get_result` (status went `queued`/`running` for ~7 min, then `failed`), resubmitted with a new idempotency key, verified the second receipt. If run 20006 is ever self-retried, memory node 2852 (from run 20013) would be duplicated — please confirm terminal-failed runs are not resurrected, or dedupe on resubmission content.

[...]

PR #864, opened 2026-08-28
fix(inbound): tell "event not found" apart from "not visible to your connection" (#854)

### LINKED ISSUE #854: illo_get_result transport failure leaves accepted preservation event unverifiable
## Observed

A Codex preservation submission was accepted and returned an event ID, but three consecutive receipt polls failed at the MCP transport before returning any result.

- Event ID: `3027cd6e-475f-426c-983a-baa353be7cf7`
- Illo run ID from submit: `19739`
- Date: 2026-08-27
- Source tool: Codex

Exact poll call:

```json
{
  "tool": "mcp__illo__illo_get_result",
  "arguments": {
    "event_id": "3027cd6e-475f-426c-983a-baa353be7cf7",
    "include_payload": false,
    "limit": 10
  }
}
```

Every attempt returned:

```text
tool call failed for `illo/illo_get_result`
Caused by:
Transport send error: Transport [rmcp::transport::worker::WorkerTransport<rmcp::transport::streamable_http_client::StreamableHttpClientWorker<codex_rmcp_client::http_client_adapter::StreamableHttpClientAdapter>>] error: Client error: HTTP request failed: http/request failed: error sending request for url (https://server-rd-1.tail44d6b6.ts.net/api/mcp)
```

The immediately preceding `illo_submit` succeeded with `status=review_required`, `operation=queued`, and the event/run IDs above.

## Expected

`illo_get_result` should return the current handling status and eventually durable preservation evidence (`evidence_status=satisfied` with non-empty `mutated_target_refs`). If the backend is temporarily unavailable, the accepted event should remain queryable after recovery.

## Impact

A completed four-PR engineering milestone cannot be claimed as archived because its mandatory durable receipt cannot be verified. The idempotency key is `codex-uwear-media-ingestion-1882-1885-20260827-preserve-1`; the preservation payload contained no credentials, secrets, customer personal data, or raw transcript.

## Existing issue search

- #830 covered `illo_submit` failing before any event ID was returned and is closed.
- #829 is a broader open outage report, but this incident differs because submission succeeded and only the receipt path is unavailable.

redawear:
## What was broken

`illo_get_result` answered the string `Inbound event not found` for **three** different situations, so a caller polling a real, completed preservation got the identical answer a never-existent UUID gets.

| situation | where | old answer |
|---|---|---|
| the row is absent | `admin.py:557-566` | `Inbound event not found` |
| the row belongs to another org | `admin.py:557-566` | `Inbound event not found` |
| the row is in **your own org**, owned by a different connection | `results.py:47-49` | `Inbound event not found` |

The third is the reported one. Event `3027cd6e-475f-426c-983a-baa353be7cf7` (run `19739`, `status: completed`) resolves fine through `run.get` and read as "not found" through `illo_get_result` — so a completed four-PR milestone could not be verified as archived, and the caller could not tell "my event is gone" from "my event is fine, you are asking from the wrong connection".

## The change

`read_inbound_submission_result` now returns a typed `InboundSubmissionResultState` — `FOUND` / `NOT_FOUND` / `NOT_VISIBLE_TO_CONNECTION` — instead of signalling through an exception message. It follows the closed-`str, Enum` shape that `RunFailureStage` / `DiagnosticValueState` established in #861 last week, rather than inventing a parallel vocabulary.

**The visibility rule is a deliberate decision, not an oversight:**

- **Absent or cross-org stays indistinguishable** and keeps answering `Inbound event not found`. Telling a cross-org caller that an id exists would leak existence across a tenant boundary, so that case is untouched on purpose.
- **Same-org / different-connection becomes distinct and machine-routable.** Same tenant, same principal, so naming it leaks nothing.
- The owning `connection_id` is **deliberately not** in the response. A caller that cannot see the event does not need the owner's id.

## Scope choice you may want to revisit

The ticket offered two acceptable routes: resolve the event across connections, **or** keep the scoping and return a distinct error. This PR takes the second.

That means the receipt is now **diagnosable** cross-connection but still not **readable** cross-connection — you learn to poll from the owning connection instead of being told "not found". I chose the conservative route because dropping the connection scope is a real isolation change that shouldn't be made unattended. If you'd rather any connection in an org read that org's inbound events, that is a small follow-up on this same seam.

## Code-quality review — one BLOCKING finding, folded

A Codex thermo-nuclear review (cross-family: the diff was Codex-written, this was a second independent Codex pass) returned **BLOCKING** on the first commit, and it was right. `InboundSubmissionResult` only *resembled* the pattern it was copying, because it could still express a contradiction:

- `owned_by_another_connection` repeated what `state` already said — one fact in two fields, free to disagree;
- `payload={}` was a placeholder standing in for "no payload", and nothing stopped a `NOT_FOUND` result carrying a full one.

Commit 2 makes the state the single owner: the boolean is gone, `payload` is `dict | None`, and `__post_init__` enforces the biconditional — **a payload is present if and only if the state is `FOUND`** — so the contradictory combinations are no longer constructible.

The review's three non-blocking notes are folded in the same commit: `_result_api()` (a helper returning two imports as an untyped positional tuple) is deleted for ordinary imports, and the four visibility tests move out of the *reconciliation* suite into a new `tests/test_inbound_results.py`, importing the seeding helpers rather than copying them.

**The wire response is unchanged by that refactor** — `_tool_get_result` still returns `owned_by_another_connection: true`, now derived as a literal in its own branch. The key is useful to an MCP caller; only the internal field was redundant.

Two things I did **not** change, so they are decisions rather than misses: the router keeps a bare `assert result.payload is not None` as a type-narrowing hint (the invariant is already enforced in `__post_init__`, so it cannot fire), and the new test module imports the `session` fixture from `test_inbound_reconciliation` instead of moving it to `conftest.py` — which avoids duplicating the fixture, at the cost of a test module importing another.

## Verification

Local only, run **serially**, never beside a live worker:

| run | result |
|---|---|
| commit 1 (`39a7f497`) | **5018 passed**, 0 failed, 11 skipped |
| commit 2 (`c0c68cc5`) | **5018 passed**, 0 failed, 11 skipped |

Count reconciled rather than assumed: `origin/main` is 5014, plus 4 new tests = 5018. The refactor moves tests rather than adding them, so the number is unchanged across the two commits — as it should be.

Command is the repo's CI lane verbatim, with **no path argument** (`pytest.ini` sets `testpaths = tests meetbot/tests`, and naming a path silently drops `meetbot/tests`):

```
venv/bin/python -m pytest -m "not requires_db and not requires_browser and not live_provider" -q
```

Frontend lane is not selected — the diff touches no `frontend/**`.

Both Codex workers self-reported `BLOCKED` / `COMMIT: none`. Both were wrong the documented way: the "block" was the sandbox `index.lock` denial (no lock file existed at either reap) and the reds were `test_gpu_integration` socket-bind errors plus one auth-callback test. Re-run serially by the orchestrator, both came back 0 failed. The orchestrator committed both.

## How to check it yourself

On staging, from an agent connection that is **not** the one that submitted an event:

1. `illo_get_result` with a **random UUID** → still `Inbound event not found`.
2. `illo_get_result` with event `3027cd6e-475f-426c-983a-baa353be7cf7` → now `state: not_visible_to_connection`, and no `connection_id` in the response.
3. `illo_get_result` from the **owning** connection → the normal receipt payload, unchanged.

The interesting one is 1 vs 2: those two used to be byte-identical.

Closes #854

### DIFF
--- a/brain/app/api/routers/agent_mcp.py
+++ b/brain/app/api/routers/agent_mcp.py
@@ -1374,8 +1374,9 @@ async def _tool_get_result(
         return {
             "event_id": event_id,
             "state": result.state.value,
-            "owned_by_another_connection": result.owned_by_another_connection,
+            "owned_by_another_connection": True,
         }
+    assert result.payload is not None
     return result.payload
 
 
--- a/brain/app/api/routers/agent_mcp.py
+++ b/brain/app/api/routers/agent_mcp.py
@@ -54,7 +54,10 @@ from brain.systems.cycles.serializers import (
 from brain.systems.cycles.service import async_run_cycle_now
 from brain.systems.external_agents import service as external_agents
 from brain.systems.inbound import admin as inbound_admin
-from brain.systems.inbound.results import read_inbound_submission_result
+from brain.systems.inbound.results import (
+    InboundSubmissionResultState,
+    read_inbound_submission_result,
+)
 from brain.systems.inbound.service import submit_inbound_envelope as _submit_inbound_envelope
 from brain.systems.knowledge.search import search_knowledge
 from brain.systems.runs.cortex.read_models import (
@@ -1365,6 +1368,14 @@ async def _tool_get_result(
     )
     if result.mutated_inbound:
         await db.commit()
+    if result.state is InboundSubmissionResultState.NOT_FOUND:
+        raise ValueError("Inbound event not found")
+    if result.state is InboundSubmissionResultState.NOT_VISIBLE_TO_CONNECTION:
+        return {
+            "event_id": event_id,
+            "state": result.state.value,
+            "owned_by_another_connection": result.owned_by_another_connection,
+        }
     return result.payload
 
 
--- a/brain/systems/inbound/results.py
+++ b/brain/systems/inbound/results.py
@@ -25,9 +25,13 @@ class InboundSubmissionResultState(str, Enum):
 @dataclass(frozen=True)
 class InboundSubmissionResult:
     state: InboundSubmissionResultState
-    payload: dict[str, Any]
-    owned_by_another_connection: bool = False
+    payload: dict[str, Any] | None = None
     mutated_inbound: bool = False
+
+    def __post_init__(self) -> None:
+        has_payload = self.payload is not None
+        if has_payload != (self.state is InboundSubmissionResultState.FOUND):
+            raise ValueError("payload must be present if and only if state is FOUND")
 
 
 def _result_handling(action_result: dict[str, Any]) -> dict[str, Any]:
@@ -63,14 +67,11 @@ async def read_inbound_submission_result(
         )
     except inbound_admin.InboundAdminError:
         return InboundSubmissionResult(
-            payload={},
             state=InboundSubmissionResultState.NOT_FOUND,
         )
     if str(event.connection_id) != str(connection_id):
         return InboundSubmissionResult(
-            payload={},
             state=InboundSubmissionResultState.NOT_VISIBLE_TO_CONNECTION,
-            owned_by_another_connection=True,
         )
 
     action_result = dict(event.action_result or {})
--- a/brain/systems/inbound/results.py
+++ b/brain/systems/inbound/results.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
+from enum import Enum
 from typing import Any
 
 from sqlalchemy.ext.asyncio import AsyncSession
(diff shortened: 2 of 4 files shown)


redawear:
> *This was generated by AI during a routine run.*

Folded into **https://github.com/Illospace/illospace/pull/866** so the three ready fixes are one merge click instead of three.

#866 is cut from current `main` (`d2978080`), carries this PR's commits unchanged, and is gated on the merged tree: **5028 passed, 0 failed** against a 5014 baseline. It also carries a third fix (#865), which shares `tests/test_inbound_reconciliation.py` with PR #864 — folding them here resolves that overlap once, rather than leaving it for the second merge.

Merge #866; this PR can then be closed. Left open for now, and it stays mergeable on its own if you prefer it that way.

redawear:
> *This was generated by AI during a routine run.*

Closing as **shipped**. [#866](https://github.com/Illospace/illospace/pull/866) merged to `main` at 2026-08-31T12:56:12Z (merge commit `33b4811b`) and carries this PRs content.

Verified by content probe on `origin/main`, not by merge status:

- `brain/systems/inbound/results.py` — `InboundSubmissionResultState` (7 occurrences)
- `brain/app/api/routers/agent_mcp.py:1373` — the `NOT_VISIBLE_TO_CONNECTION` branch
- `tests/test_inbound_results.py` present on `main`

Control probes ran both ways so an empty result could not read as a clean one. [#854](https://github.com/Illospace/illospace/issues/854) auto-closed on the same merge.

Issue #874, opened 2026-09-01
illo_submit returns 502 before event ID is issued


axel-havard:
## Summary

`illo_submit` returned an HTTP 502 Bad Gateway before producing an event ID while handling a `preserve_knowledge` request from Codex.

## Exact tool call

```text
mcp__illo__illo_submit({
  source_tool: "codex",
  origin: "codex.session-preserve",
  desired_outcome: "preserve_knowledge",
  idempotency_key: "customer-spend-analysis-2026-09-01-preserve-1",
  session_id: "customer-spend-analysis-2026-09-01",
  repo: "/Users/axelhavard/Documents/Repos/app",
  task_title: "Customer spend cohort analysis",
  constraints: { privacy: "Do not store personal identifiers or raw customer-level records." },
  message: "Preserve aggregate conclusions and a durable local report reference from a completed customer cohort analysis."
})
```

The original message contained only aggregate business metrics and a local artifact path; this public issue intentionally omits those details.

## Observed behavior

```text
Transport send error ... unexpected server response: HTTP 502: 502 Bad Gateway (nginx/1.27.5)
```

No `event_id` was returned, so the caller could not poll `illo_get_result` or verify durable evidence.

## Expected behavior

The submission should return an async event ID, or a structured retriable error with enough information to safely retry the same idempotent request.

## Context

- Source tool: Codex
- Date: 2026-09-01
- Desired outcome: `preserve_knowledge`
- Event ID: unavailable because the gateway failed before acknowledgement

redawear:
> *This was generated by AI during an autonomous routine run (R3).*

## Root cause: the server root filesystem is 100% full. Postgres is in a crash-restart loop it cannot exit.

The 502 is the last link in a five-step chain, not the fault. Measured first-hand on `illo-dev` (`server-rd-1`) at 2026-09-01 16:47–16:55Z.

```
/dev/mapper/ubuntu--vg-ubuntu--lv  1.8T  1.8T  0  100%  /
```

### The chain

| when (UTC) | what | evidence |
|---|---|---|
| 15:05:52 | Disk hits 100%. Postgres cannot extend a relation. | `ERROR: could not extend file "base/16384/18095": No space left on device` |
| 15:10:26 | First `PANIC`. Postgres begins crash-looping. | `PANIC: could not write to file "pg_logical/replorigin_checkpoint.tmp": No space left on device` |
[...]

Issue #780, opened 2026-08-10
[runtime] Capacity sensor — Illo measures its own disk on a schedule and records the trend


redawear:
> *This was generated by AI during triage.*

Slice **2 of 3** of [#752](https://github.com/Illospace/illospace/issues/752),
sliced on Reda's instruction (2026-08-10). Read #752 first for the evidence and
the prior art.

## Problem

Illo filled its own disk to 98% and never noticed. A human had to look. On
2026-08-09 the illo-dev root filesystem reached 1.7T of 1.8T with 49G free, and
Illo ran normally throughout, because it has no way to know.

The only `shutil.disk_usage` call in the codebase is `_get_disk_info` at
`brain/app/api/routers/system.py:654-665`, surfaced on `GET /system` at `:826`.
No tool in the catalog reads it. There is no capacity tool and no
workspace-inventory tool.

There is also no durable record of any capacity measurement, so there is no
trend, no threshold edge, and no "it has been degrading for three days".

## Scope

1. A scheduled job that measures the filesystem and the workspace volume, and
   writes one `memory_health_log` row per run with `check_type="host_capacity"`
   and `details` = `{mount, pct_used, bytes_free, top_consumers[]}`.
   `nightly_memory_health` (`brain/jobs/pipelines/nightly_memory_health.py:21-50`,
   `brain/platform/db/models/memory_health.py:27-46`) already has this shape — it
   fits without a new table.
2. A tool that answers "how much disk do I have, and what is using it", so Illo
   can read capacity without raw `exec_command`.
3. Crossing the warn threshold reaches a human **once**, through the existing
   alert path with a latch — `scheduler_alert_latches`
   (`brain/platform/db/models/scheduler.py:44-68`), the neutral failure-guard
   engine, and Slack delivery with the channel as a policy field.

## What must not happen

Cadence belongs in the scheduler catalog. **What counts as healthy comes from the
storage policy row, not a constant** — see the sibling slice. A threshold
hardcoded here fails the parent ticket's one rule.

The alarm must latch. #666/#675 is the anchor: an unlatched alarm on an abandoned
metric paged Slack until "illo is choking on slack".

## Acceptance

- A capacity measurement is recorded on a schedule and readable as a trend.
- Illo can answer "how much disk do I have, and what is using it" through a tool.
- Crossing the warn threshold notifies a human once, not on every tick.
- No threshold constant is introduced in this slice.

## Sequencing

Depends on the storage policy row slice — it reads the thresholds from there.

[...]

Issue #877, opened 2026-09-01
[runs][project-context] Sibling headless workers each re-clone the same project context — 6 byte-identical workspaces per run, ~6x the disk


redawear:
> *This was generated by AI during triage.*

## Problem

Every headless worker spawned by one parent run clones the same project context into its own workspace, so a single run leaves **six byte-identical ~1 GB trees** instead of one. This multiplies workspace disk by roughly 6x and makes the ~38 GB/day growth rate about six times larger than the work requires.

This was **cause 2 of [#750](https://github.com/Illospace/illospace/issues/750)** ("one workspace per spawned WORKER, not per run — 851 dirs for 140 parent runs"). #750 was closed on the strength of a manual cleanup plus [#752](https://github.com/Illospace/illospace/issues/752), and #752's three slices ([#779](https://github.com/Illospace/illospace/issues/779) / [#780](https://github.com/Illospace/illospace/issues/780) / [#781](https://github.com/Illospace/illospace/issues/781)) built retention, sensing and reclamation — they made the garbage collectable, not smaller. The duplication itself was never fixed and has had no open ticket since.

## Evidence

From the `workspace_gc` inventory on `illo-dev`, 2026-09-01 18:07 UTC, immediately after the disk-full outage ([#874](https://github.com/Illospace/illospace/issues/874)). Grouping the 30 surviving workspaces by parent run:

| parent run | workspaces | bytes each | total |
|---|---|---|---|
| 20879 | 6 | 1,048,792,460 | ~6.3 GB |
| 20899 | 6 | 1,048,792,460–506 | ~6.3 GB |
| 20928 | 6 | 93,177,336 | ~559 MB |
| 20972 | 6 | 29–52 MB | ~274 MB |
| 21008 (running) | 6 | 110–150 MB | ~776 MB |

Five runs, six workspaces each, no exceptions. The byte sizes within a run are identical to the byte — for 20928 all six are exactly 93,177,336 bytes. These are not workspaces that diverged; they are the same clone six times.

Sample paths:

```text
ideas/headless-worker-20879-1034836dc6d1d33a   1048792460 bytes
ideas/headless-worker-20879-1c849d76f809c881   1048792460 bytes
ideas/headless-worker-20879-47af232ce49560d3   1048792460 bytes
ideas/headless-worker-20879-731cff4c33a63730   1048792460 bytes
ideas/headless-worker-20879-c16a7501dfdebf29   1048792460 bytes
[...]

Issue #876, opened 2026-09-01
[runtime] Workspace reclamation shipped disabled — the seed policy row sets automatic_reclamation_allowed=FALSE, so the reclaim lever has skipped every run since 2026-08-11


redawear:
> *This was generated by AI during triage.*

## Problem

The workspace reclaim lever from [#781](https://github.com/Illospace/illospace/issues/781) has never reclaimed a single byte in production. Migration `0060_storage_policies` seeds the only policy row with `automatic_reclamation_allowed = FALSE`, and nothing has changed it since it ran on 2026-08-11. The `workspace_gc` scheduler job has therefore run on schedule, found the disk filling, and refused to act — every hour, for three weeks.

On 2026-09-01 that ended in a full outage: the `illo-dev` root filesystem reached 100%, postgres crash-looped on `PANIC: could not write to file "pg_logical/replorigin_checkpoint.tmp": No space left on device`, `migrate` failed with `CannotConnectNowError`, and `api` / `scheduler` / `worker` never started. That is the outage behind [#874](https://github.com/Illospace/illospace/issues/874).

This is the second half of the same failure as [#875](https://github.com/Illospace/illospace/issues/875). #875 explains why nobody was told. This issue explains why nothing acted.

## Evidence

The `workspace_gc` job at 14:47 UTC, ~2h20m before the stack died:

```json
{"action": "reclaim", "automatic": true, "automatic_reclamation_allowed": false,
 "bytes_reclaimable": 0, "bytes_reclaimed": 0, "directories_scanned": 0,
 "reclamation_skipped_reason": "automatic_reclamation_disabled_by_policy",
 "retention_hours": 48, "workspaces": []}
```

```text
2026-09-01 14:47:12,989 [workspace_gc] Workspace GC skipped because automatic reclamation is disabled
```

The `host_capacity` job at 15:07 UTC, in the same run, naming the exact directory that was killing the box:

```json
{"status": "critical", "alert_sent": false,
 "details": {"mount": "/workspaces", "pct_used": 95.9, "bytes_free": 0,
             "top_consumers": [{"path": "ideas", "bytes_used": 572805518201}, ...]},
 "thresholds": {"warn_percent": 80, "critical_percent": 90}}
```

[...]

PR #878, opened 2026-09-01
fix(runtime): let the host-capacity alert escalate from warn to critical

### LINKED ISSUE #875: [runtime] Host-capacity alert cannot escalate — warn and critical share one latch, so the critical crossing is always silent

## Problem

The host-capacity guard alerts **once**, at the warning threshold, and can never alert again until storage returns to healthy. Crossing the *critical* threshold produces no alert at all. So the alarm is loudest when the problem is smallest, and silent when the disk actually fills.

This is not theoretical. On 2026-09-01 the `illo-dev` root filesystem reached 100%, postgres entered a crash-restart loop, `migrate` failed, and `api`/`scheduler`/`worker` never started — the outage in [#874](https://github.com/Illospace/illospace/issues/874). Nobody was alerted. A human found it by hand ~19 minutes after the API died.

## Cause

`brain/systems/host_capacity.py` computes three states but arms only one latch.

`_capacity_status()` returns `ok` / `warn` / `critical`:

```python
def _capacity_status(pct_used, policy) -> str:
    if pct_used >= policy.capacity_critical_percent:
        return "critical"
    if pct_used >= policy.capacity_warn_percent:
        return "warn"
    return "ok"
```

But `record_host_capacity()` collapses both non-ok states into a single trigger with a single key:

```python
HOST_CAPACITY_ALERT_KEY = "host_capacity_warn"
_CAPACITY_WARN_TRIGGER_KIND = FailureGuardTriggerKind(HOST_CAPACITY_ALERT_KEY)
...
trigger = FailureGuardTriggerResult(
    kind=_CAPACITY_WARN_TRIGGER_KIND,
    active=status != "ok",     # <- warn and critical are indistinguishable here
    ...
)
```

And the guard only fires on an unlatched edge — `brain/systems/failure_guard/core.py:439-441`:

```python
crossed = (
    new_edge_mode != "ignore" and result.active and latch is None
)
```

Once the warn alert latches `host_capacity_warn`, `latch is None` is false on every later tick. The latch is deleted only when `status == "ok"`. Therefore:

| tick | pct used | status | alert |
|---|---|---|---|
| 1 | 80% | warn | ✅ fires, latch set |
| 2 | 92% | **critical** | ❌ silent — latch exists |
| 3 | 100% | **critical** | ❌ silent — latch exists |

The `alert_title` field already distinguishes the two (`"Host capacity critical"` vs `"Host capacity warning"`), which shows the escalation was intended. Only the latch key and trigger kind were never split, so the critical title can only ever be rendered on a tick that is also the *first* non-ok tick.

## Why it stayed hidden

`tests/test_host_capacity.py:172` is
`test_host_capacity_warn_alert_latch_fires_exactly_once_across_two_over_threshold_ticks`.
Both of its ticks are over the **warn** threshold, so it pins the once-only behavior as correct — which it is, for a sustained warning. There is no test where a warn tick is followed by a critical tick, so the escalation gap is uncovered rather than accepted.

## Expected behavior

Crossing into `critical` is its own edge and alerts once, even when a warning already latched. A sustained critical state must still not re-alert every hour — the latch discipline is right, there just needs to be one latch per severity.

## Acceptance criteria

1. `warn` and `critical` own separate latch keys (e.g. `host_capacity_warn` and `host_capacity_critical`), so a warn→critical transition delivers exactly one new alert carrying the `"Host capacity critical"` title.
2. A second consecutive `critical` tick delivers **no** alert.
3. Returning to `ok` clears **both** latches, so a later warn or critical crossing alerts again.
4. Falling from `critical` back to `warn` does not re-alert (the warn latch is still held) and does not clear the critical latch prematurely — going critical again after that must alert.
5. New tests cover the warn→critical transition and each case above. The existing test at `tests/test_host_capacity.py:172` keeps passing unchanged.
6. `SchedulerAlertLatch` rows are keyed by string, so no migration should be needed — confirm that and say so in the PR.

## Out of scope

The sensor's `top_consumers` only inventories Illo's own workspace root, so in [#874](https://github.com/Illospace/illospace/issues/874) it would have reported 100% used while listing consumers summing to a few GB — the 1.1T that actually filled the disk sits outside the workspace and is invisible to it. That is a separate defect; do not change it here.

## Files

- `brain/systems/host_capacity.py` — `HOST_CAPACITY_ALERT_KEY`, `_CAPACITY_WARN_TRIGGER_KIND`, `record_host_capacity`, `_HostCapacityLatchStore`
- `brain/systems/failure_guard/core.py:421-459` — `async_evaluate_failure_edges` (read-only; the edge rule is correct, the caller's key is not)
- `tests/test_host_capacity.py`

## Prior art

[#752](https://github.com/Illospace/illospace/issues/752) (parent, closed) → [#779](https://github.com/Illospace/illospace/issues/779) storage policy row, [#780](https://github.com/Illospace/illospace/issues/780) capacity sensor, [#781](https://github.com/Illospace/illospace/issues/781) reclaim lever. #780 shipped the latched warn alert; this issue is the escalation half that was never specified.

redawear:
Closes #875

## What was broken

`_capacity_status()` has returned three states — `ok` / `warn` / `critical` — since [#780](https://github.com/Illospace/illospace/issues/780) shipped the sensor. But both non-ok states shared **one** latch key, and the failure guard only fires on an unlatched edge (`brain/systems/failure_guard/core.py:439-441`):

```python
crossed = (new_edge_mode != "ignore" and result.active and latch is None)
```

So once the warning alerted, every later tick was silent — including the critical crossing, and including 100% full. The latch cleared only on a return to `ok`.

The alarm was loudest when the problem was smallest, and mute when the disk actually filled.

This is not hypothetical. It is why [#874](https://github.com/Illospace/illospace/issues/874) happened with no warning: the root filesystem hit 100%, postgres entered a crash-restart loop, and the API never started. Nobody was paged. A human found it by hand.

## What this does

Each severity gets its own latch, and everything about a severity comes from one ordered table:

```python
_CAPACITY_SEVERITIES = (
    _CapacitySeverity(status="warn",     alert_key=HOST_CAPACITY_ALERT_KEY,          priority=1, ...),
    _CapacitySeverity(status="critical", alert_key=HOST_CAPACITY_CRITICAL_ALERT_KEY, priority=2, ...),
)
```

Which latches to clear, which triggers are active, which edge wins, the alert title, the severity word and the threshold percent all derive from it. Both are priority comparisons against the current status, so a third severity would need no new branches.

**One alert per tick.** When a jump crosses both thresholds at once, only the highest is delivered — critical subsumes warning — but *both* edges are latched, so the warning cannot fire later on the way down.

## Behavior, as pinned by tests

| tick | status | alert delivered | latches after |
|---|---|---|---|
| cold start at 95% | critical | **1** — "Host capacity critical" | warn + critical |
| again at 95% | critical | none | warn + critical |
| 80% | warn | none | warn |
| 95% again | critical | **1** — "Host capacity critical" | warn + critical |
| 50% | ok | none | none |

`warn → critical` delivers exactly one critical alert. A sustained state never re-alerts.

## Verification — read this before merging

**The full suite did not run, and I will not claim it did.** The host sat at load ~50 for this entire run from unrelated work. The guardrail is load > 20 invalidates a suite result in either direction, so a green number produced there would have been decoration. I started the repo's CI command verbatim, gave it 45 minutes, and stopped it at ~10% of a core rather than report a number that proves nothing.

What did run, serially, on the final commit:

```
pytest -m "not requires_db and not requires_browser and not live_provider" -q \
  tests/test_host_capacity.py tests/test_scheduler.py \
  tests/test_scheduler_programs.py tests/test_tool_registry.py
→ 115 passed in 295.72s
```

Those four files are the complete set that references `host_capacity` anywhere in the repo, so this is the change's full blast radius — but it is a subset of CI, not CI.

The diff touches no `frontend/**` path, so that lane is not selected.

**This cannot be exercised on the live host until [#874](https://github.com/Illospace/illospace/issues/874) is fixed.** The scheduler container is in `Created` and never started, so the hourly `host_capacity` job is not running at all right now.

## Review pass

A Codex code-quality review of the first commit (`3e70efab`) returned **BLOCKING** on three points. All three are folded into `27519810`:

1. the ok→critical double alert (it independently recommended suppressing the warning);
2. the lifecycle smeared across four places;
3. `replace(guard, edges=(edge,))` synthesizing a partial aggregate to reach around `async_latch_failure_edges`.

## Production receipt — this defect is recorded, not just reasoned about

The hourly sensor ran at 15:07Z on 2026-09-01, 18 minutes before the API died, and wrote this:

```json
{"status": "critical", "pct_used": 95.9, "bytes_free": 0, "mount": "/workspaces",
 "alert_sent": false,
 "thresholds": {"warn_percent": 80, "critical_percent": 90}}
```

`status: critical`, **zero bytes free**, `alert_sent: false`. The warning latch was already held, so the critical crossing produced nothing — while postgres was minutes from a crash-loop that took the API down for 2.5 hours ([#874](https://github.com/Illospace/illospace/issues/874)). That is exactly the branch this PR changes.

## Correction to an earlier claim on #875

The issue body says the sensor's `top_consumers` "would have reported 100% used while listing consumers summing to a few GB — pointing away from the 1.1T that actually filled the disk." **That was wrong and I have withdrawn it.** I inferred it from the code rather than from what the sensor recorded; the real 15:07Z row named `ideas` at **572.8 GB**, the largest reclaimable object on the box.

What survives is narrower: `mount` is `/workspaces`, not `/`, so the sensor read 95.9% on its own volume while the root filesystem was at 100%. Pressure originating outside that volume is invisible to it. Still a separate change from this one, which is only about whether the alarm fires at all.

No migration: `SchedulerAlertLatch` is keyed by an `alert_key` string, so the new severity is a new row.

### DIFF
--- a/brain/systems/host_capacity.py
+++ b/brain/systems/host_capacity.py
@@ -5,7 +5,7 @@ import logging
 import os
 import shutil
 from collections.abc import Awaitable, Callable
-from dataclasses import dataclass, replace
+from dataclasses import dataclass
 from datetime import datetime
 from pathlib import Path
 from typing import Any, TypedDict
@@ -45,14 +45,65 @@ logger = logging.getLogger(__name__)
 HOST_CAPACITY_CHECK_TYPE = "host_capacity"
 HOST_CAPACITY_ALERT_KEY = "host_capacity_warn"
 HOST_CAPACITY_CRITICAL_ALERT_KEY = "host_capacity_critical"
-_CAPACITY_WARN_TRIGGER_KIND = FailureGuardTriggerKind(HOST_CAPACITY_ALERT_KEY)
-_CAPACITY_CRITICAL_TRIGGER_KIND = FailureGuardTriggerKind(
-    HOST_CAPACITY_CRITICAL_ALERT_KEY
+
+
+@dataclass(frozen=True, slots=True)
+class _CapacitySeverity:
+    status: str
+    alert_key: str
+    trigger_kind: FailureGuardTriggerKind
+    alert_title: str
+    error_severity: str
+    priority: int
+    threshold_percent: Callable[[StoragePolicyValues], int]
+
+
+_CAPACITY_SEVERITIES = (
+    _CapacitySeverity(
+        status="warn",
+        alert_key=HOST_CAPACITY_ALERT_KEY,
+        trigger_kind=FailureGuardTriggerKind(HOST_CAPACITY_ALERT_KEY),
+        alert_title="Host capacity warning",
+        error_severity="warning",
+        priority=1,
+        threshold_percent=lambda policy: policy.capacity_warn_percent,
+    ),
+    _CapacitySeverity(
+        status="critical",
+        alert_key=HOST_CAPACITY_CRITICAL_ALERT_KEY,
+        trigger_kind=FailureGuardTriggerKind(HOST_CAPACITY_CRITICAL_ALERT_KEY),
+        alert_title="Host capacity critical",
+        error_severity="critical",
+        priority=2,
+        threshold_percent=lambda policy: policy.capacity_critical_percent,
+    ),
 )
-_CAPACITY_ALERT_KEY_BY_TRIGGER_KIND = {
-    _CAPACITY_WARN_TRIGGER_KIND: HOST_CAPACITY_ALERT_KEY,
-    _CAPACITY_CRITICAL_TRIGGER_KIND: HOST_CAPACITY_CRITICAL_ALERT_KEY,
-}
+
+
+def _capacity_severity_for_kind(
+    trigger_kind: FailureGuardTriggerKind,
+) -> _CapacitySeverity:
+    severity = next(
+        (
+            severity
+            for severity in _CAPACITY_SEVERITIES
+            if severity.trigger_kind == trigger_kind
+        ),
+        None,
+    )
+    if severity is None:
+        raise ValueError(f"Unsupported host-capacity trigger: {trigger_kind}")
+    return severity
+
+
+def _capacity_priority(status: str) -> int:
+    if status == "ok":
+        return 0
+    return next(
+        severity.priority
+        for severity in _CAPACITY_SEVERITIES
+        if severity.status == status
+    )
 
 
 class HostCapacityDetails(TypedDict):
@@ -107,7 +158,10 @@ class _HostCapacityLatchStore:
         latches = await self.session.scalars(
             select(SchedulerAlertLatch).where(
                 SchedulerAlertLatch.alert_key.in_(
-                    _CAPACITY_ALERT_KEY_BY_TRIGGER_KIND.values()
+                    tuple(
+                        severity.alert_key
+                        for severity in _CAPACITY_SEVERITIES
+                    )
                 )
             )
         )
@@ -121,11 +175,9 @@ class _HostCapacityLatchStore:
         trigger_kind: FailureGuardTriggerKind,
         alerted_at: datetime,
     ) -> SchedulerAlertLatch:
-        alert_key = _CAPACITY_ALERT_KEY_BY_TRIGGER_KIND.get(trigger_kind)
-        if alert_key is None:
-            raise ValueError(f"Unsupported host-capacity trigger: {trigger_kind}")
+        severity = _capacity_severity_for_kind(trigger_kind)
         latch = SchedulerAlertLatch(
-            alert_key=alert_key,
+            alert_key=severity.alert_key,
             alerted_at=alerted_at,
         )
         self.session.add(latch)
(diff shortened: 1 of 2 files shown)


redawear:
Superseded by #879, which carries this change plus #876 and #877 — the other two halves of the 2026-09-01 disk-full outage.

#876 rewrites `brain/systems/host_capacity.py` on top of this branch (it adds a third alert for "capacity is unhealthy while reclamation is policy-disabled"), so merging the two separately would mean resolving that overlap by hand. #879 is one merge click for all three.

A maintainability review of the combined branch then split `_CapacitySeverity` into a `_CapacityAlert` base and an ordered `_CapacityLevel` ladder, so the escalation fix in #879 reads differently from the one here — same behaviour, and the acceptance checks on #875 are unchanged.

Leaving this open until #879 merges.

redawear:

redawear:

Verified on `origin/main` by content probe, not by merge status: `HOST_CAPACITY_CRITICAL_ALERT_KEY` is present at `brain/systems/host_capacity.py:48`. Closing this original now that its content has shipped.

PR #879, opened 2026-09-01
fix(runtime): the four fixes behind the 2026-09-01 disk-full outage — alert escalation, reclamation enabled, one context clone per run, and the unwatched root filesystem

### LINKED ISSUE #875: [runtime] Host-capacity alert cannot escalate — warn and critical share one latch, so the critical crossing is always silent

### LINKED ISSUE #876: [runtime] Workspace reclamation shipped disabled — the seed policy row sets automatic_reclamation_allowed=FALSE, so the reclaim lever has skipped every run since 2026-08-11

### LINKED ISSUE #877: [runs][project-context] Sibling headless workers each re-clone the same project context — 6 byte-identical workspaces per run, ~6x the disk

### LINKED ISSUE #880: [runtime] Capacity sensor watches only the workspace volume — the root filesystem that caused the 2026-09-01 outage is unmonitored

Found by R3 while re-deriving the 2026-09-01 disk-full outage. Independent of
[#875](https://github.com/Illospace/illospace/issues/875),
[#876](https://github.com/Illospace/illospace/issues/876) and
[#877](https://github.com/Illospace/illospace/issues/877): those three are all inside
draft PR [#879](https://github.com/Illospace/illospace/pull/879), and that PR does not
change the behaviour below.

## Problem

The capacity sensor measures **one** filesystem — the workspace volume — so it cannot
observe the filesystem whose exhaustion actually takes Illo down.

`brain/jobs/pipelines/host_capacity.py:17-20` passes a single path:

```python
return await record_host_capacity(
    uow.session,
    workspace_root=brain_config.resolve_workspace_root(),
)
```

`measure_host_disk_capacity` (`brain/systems/host_capacity.py:206-219`) then calls
`shutil.disk_usage(root)` on that one path and walks up to its mount point. One path
in, one `mount` / `pct_used` / `bytes_free` row out.

In production that path is a **separate volume from root**:

- `deploy/compose/docker-compose.yml:33` — `WORKSPACE_ROOT: /workspaces`
- `deploy/docker/api.Dockerfile:11` — `WORKSPACE_ROOT=/workspaces`
- `/workspaces` is the `illospace_illo_workspaces` docker volume.

## Evidence that this is the gap that mattered

On 2026-09-01 the `illo-dev` **root** filesystem hit 100% at 15:05Z. Postgres PANICked
on its end-of-recovery checkpoint and entered a crash-restart loop, `illospace-migrate-1`
exited 1 with `CannotConnectNowError`, so `api`, `scheduler` and `worker` never started
and nginx returned 502 for ~2.5 hours.

The sensor ran at 15:07Z, two minutes into the outage, and recorded **95.9% used on
`/workspaces`** — a warn-level reading, while the filesystem that had actually failed was
full. The 81 GB Docker build cache that had to be reclaimed to restore service also lives
on root, not on `/workspaces`.

So the sensor was not merely latched silent (#875) — even with a perfect latch it was
reading a filesystem that was not the one failing.

## This is a regression against the sensor's own ticket

[#780](https://github.com/Illospace/illospace/issues/780) is the ticket that built this
sensor. Its Problem section is explicitly about the root filesystem:

> On 2026-08-09 the illo-dev root filesystem reached 1.7T of 1.8T with 49G free

and its scope item 1 asks for a job that measures

> the filesystem **and** the workspace volume

Two mounts were specified. One shipped. Single-mount is therefore not a deliberate design
choice to re-litigate — it is the specified scope, unfinished.

## Change

Measure every filesystem the runtime depends on, not just the workspace volume, and
threshold each one independently.

1. Give the sensor a set of monitored paths rather than a single `workspace_root`.
   The root filesystem (`/`) and the workspace volume are the minimum; keep the existing
   `WORKSPACE_ROOT` contract for the workspace entry.
2. De-duplicate by resolved mount point, so a deploy where both paths sit on one
   filesystem records one row and alerts once, not twice.
3. Record one `memory_health_log` row per **mount** with the existing
   `check_type="host_capacity"` shape, so the trend stays per-filesystem.
4. Threshold and latch per mount — the alert key must include the mount, or a warn on one
   filesystem will silence a critical on another. This is the same latch-collision class as
   #875, one dimension over; #879 fixes the warn/critical collision but leaves every mount
   sharing a key.
5. Keep the workspace inventory (`top_consumers`) scoped to the workspace volume, where a
   directory walk is meaningful. A root-filesystem row may carry disk numbers with no
   inventory rather than walking `/`.

## Acceptance

- The sensor records a row for the root filesystem and a row for the workspace volume on
  a deploy where they are different mounts.
- Replaying the 2026-09-01 conditions — root 100%, `/workspaces` 95.9% — produces a
  **critical** row and one alert for the root filesystem.
- Two paths that resolve to the same mount produce exactly one row and one alert.
- A critical on one mount still alerts when another mount is already latched at warn.
- `top_consumers` stays populated for the workspace volume.

## Not in scope

Changing thresholds or the storage policy row. Thresholds keep coming from the policy row
per #779 — this ticket changes *what is measured*, not *what counts as healthy*.

redawear:
One PR for the four tickets that come out of the 2026-09-01 disk-full outage
(#874). They are folded together because Reda merges one PR per repo, and because
#875, #876 and #880 all change `brain/systems/host_capacity.py` — merging them
separately would make him resolve that overlap by hand.

Closes #875
Closes #876
Closes #877
Closes #880

## What each ticket changes

**#875 — the capacity alert could not escalate.** `warn` and `critical` shared one
latch key, so once the warn alert latched, the critical crossing was silent
forever. Each severity now owns its own latch, driven by an ordered
`_CAPACITY_SEVERITIES` table in `brain/systems/host_capacity.py`.

**#876 — reclamation shipped disabled.** Migration `0060` seeded the only policy
row with `automatic_reclamation_allowed = FALSE`, so `workspace_gc` refused to act
every hour for three weeks. Three parts:
- New migration `0065` is the single owner of the fix. It flips only rows that are
  `is_active` **and** still carry the untouched seed rationale, so a row an
  operator deliberately edited is left alone. Because the chain runs
  `0060 → … → 0064 → 0065`, this covers **fresh** installs as well as already
  migrated ones — `0060` itself is left exactly as it shipped.
- `workspace_reclamation.py` now scans **before** it refuses, so a disabled run
  still reports `directories_scanned` and `bytes_reclaimable` instead of `0`.
- A third severity row alerts when capacity is unhealthy while reclamation is
  policy-disabled — the combination that produced this outage.

**#877 — six identical clones per run.** A sibling worker re-cloned the parent's
~1 GB context because `should_use_existing_resource_path()` refused any managed
path outside its own workspace. It now takes a **required** keyword-only `access`
argument: `READ_ONLY` callers reuse a sibling's tree, `WRITE` callers still
re-clone. Because the argument is required, no future call site can inherit the
permissive branch by accident, and both WRITE sites stay re-gated by
`path_is_relative_to()`, so a foreign path can never become a write target.

**#880 — the sensor was watching the wrong filesystem.** This is the one that makes
the other three matter. The sensor measured a single path, the workspace volume, so
the **root** filesystem — the one that actually filled and took Illo down — was never
observed. On 2026-09-01 it recorded **95.9% used on `/workspaces`** while `/` sat at
**100%**. Production splits the two (`deploy/compose/docker-compose.yml:33` and
`deploy/docker/api.Dockerfile:11` both set `WORKSPACE_ROOT=/workspaces`), and the 81 GB
Docker build cache that had to be reclaimed to restore service lives on root as well.

Without this, #875 ships an alert that can now escalate correctly — pointed at a
filesystem that was not the one failing.

It is unfinished scope rather than a design choice: #780, the ticket that built this
sensor, opens on the root filesystem filling and its scope item 1 asks for a job that
measures "the filesystem **and** the workspace volume". Two mounts were specified, one
shipped. The change measures a set of paths, de-duplicates them by **resolved mount
point** (so a dev machine where both live on one filesystem still records one row and
alerts once), records one `memory_health_log` row per mount in the existing
`check_type="host_capacity"` shape, and scopes each latch key to its mount — the #875
collision class one dimension over, since every mount would otherwise share a key.
`top_consumers` stays scoped to the workspace volume, where walking the tree means
something. Thresholds still come from the storage policy row; no new migration.

## #875 acceptance item 6 — confirmed, no migration needed

`SchedulerAlertLatch.alert_key` is `mapped_column(String(80), primary_key=True)`
(`brain/platform/db/models/scheduler.py:49`). Latch keys are free-form strings, so
`host_capacity_critical` and `host_capacity_reclamation_disabled` need no schema
change. The longest new key is 34 characters.

## Verification

CI is not the evidence here. What was run, locally and serially on a quiet machine:

- The repo's own CI command **verbatim** from `.github/workflows/brain-ci.yml`,
  with no path argument (a path argument overrides `pytest.ini`'s
  `testpaths = tests meetbot/tests` and silently drops 45 tests).
- The diff touches no `frontend/**`, so the frontend lane is not selected.
- Alembic: `0064` is the head on `origin/main` and nothing else claims it as a
  parent, so `0065` is exactly one new head, correctly chained.

Two regressions were found this way and fixed in this branch rather than shipped:
the `0065` migration was missing the repo's structured docstring header, and the
new `tests/test_storage_policy.py` used sync SQLAlchemy, which broke the
async-DB debt metric. Both were proved to be real by re-running the same two test
files against a clean `origin/main` worktree, where they pass 52/52.

## How to test this on illo-dev

The change is a runtime/scheduler behaviour, so there is no screen to click. Two
job runs show all of it:

1. Run the `workspace_gc` job. Before this PR its JSON says
   `"directories_scanned": 0` with
   `"reclamation_skipped_reason": "automatic_reclamation_disabled_by_policy"`.
   After, it should report a non-zero `directories_scanned` and actually reclaim
   `headless-worker-*` directories older than the 48h retention window.
2. Run the `host_capacity` job while usage is above the warn threshold, twice.
   The first tick alerts; the second does not. Push usage past the critical
   threshold and it must alert **again** — that second alert is the thing that
   did not happen on 2026-09-01.
3. For #877, start one parent run and count the workspaces it leaves under
   `ideas/`. It should leave one materialized tree, not six byte-identical ones.

## Review

A maintainability review of this branch returned **BLOCKING** with 14 findings.
Twelve were folded into the branch before this PR opened (commit `557a46a6`); the
full review is kept at
`uwear-routines/state/evidence/875-876-877-thermo-review-20260901T2105.md`.

The three blocking clusters and what changed:

- **`_CapacitySeverity` was two types wearing one name.** The reclamation-risk row
  had no status and no threshold, and the table was only half-adopted because
  `_capacity_status()` still hardcoded the threshold ladder. Split into
  `_CapacityAlert` and an ordered `_CapacityLevel` ladder; `priority`,
  `minimum_capacity_priority`, `requires_reclamation_disabled` and both nullable
  fields are gone, and alert precedence is now stated explicitly instead of
  encoded in an unexplained integer.
- **`0060` and `0065` were two owners of one fact.** The `0060` edit is reverted
  byte-for-byte; `0065` alone now owns it.
- **The new migration tests used a bespoke fake `op` that string-sniffed SQL.**
  Replaced with the canonical `Operations(MigrationContext.configure(connection))`
  pattern that six sibling migration tests already use, in a new
  `tests/test_storage_policy_migration.py`.

## Two findings deliberately NOT folded — your call

1. **The `access` enum in #877 could be two named functions instead**
   (`existing_readable_path` / `existing_writable_path`). The reviewer marks this
   non-blocking. It was left alone because the property that matters — a required
   argument forcing every future caller to decide read vs write — is exactly what
   #877's guard exists to protect, and both spellings preserve it. This is a taste
   call, not a defect.
2. **`tests/test_host_capacity.py` grew 373 → 680 lines** with six tests that
   repeat the same four-line setup. A `_tick` helper would shrink them. Cosmetic.

## Two things to know before merging

- **Two migration-test assertions changed, both forced.** Reverting the `0060`
  edit means the seed is `FALSE` again, so the test asserting `TRUE` had to
  follow; and the real `StoragePolicy.__table__` carries a one-active-row unique
  index, so the old fixture's two active rows could not survive the move to the
  real table. The cases were split across two tests, not dropped.
- **The new #877 isolation test locks the behaviour, not the enum.** Both `WRITE`
  call sites are followed by a `path_is_relative_to()` check, so flipping one to
  `READ_ONLY` is behaviour-neutral at the materializer level and the test would
  not catch it. It does prove the child re-clones and never touches the sibling.

## Supersedes

This replaces draft PR #878, which carried #875 alone. #878 stays open until this
merges.

### DIFF
--- a/brain/jobs/pipelines/host_capacity.py
+++ b/brain/jobs/pipelines/host_capacity.py
@@ -18,7 +18,6 @@ async def run() -> dict[str, Any]:
         return await record_host_capacity(
             uow.session,  # type: ignore[arg-type]
             workspace_root=workspace_root,
-            monitored_paths={"/", workspace_root},
         )
 
 
--- a/brain/jobs/pipelines/host_capacity.py
+++ b/brain/jobs/pipelines/host_capacity.py
@@ -13,10 +13,12 @@ from brain.systems.host_capacity import record_host_capacity
 
 
 async def run() -> dict[str, Any]:
+    workspace_root = brain_config.resolve_workspace_root()
     async with UnitOfWork() as uow:
         return await record_host_capacity(
             uow.session,  # type: ignore[arg-type]
-            workspace_root=brain_config.resolve_workspace_root(),
+            workspace_root=workspace_root,
+            monitored_paths={"/", workspace_root},
         )
 
 
--- a/brain/platform/db/alembic/versions/0060_storage_policies.py
+++ b/brain/platform/db/alembic/versions/0060_storage_policies.py
@@ -124,8 +124,6 @@ def upgrade() -> None:
     active_count = op.get_bind().execute(
         sa.text("SELECT count(*) FROM storage_policies WHERE is_active = TRUE")
     ).scalar_one()
-    # This guarded seed only runs for fresh databases. Changing its defaults
-    # does not mutate deployments where this migration has already run.
     if active_count == 0:
         op.execute(
             sa.text(
@@ -146,8 +144,8 @@ def upgrade() -> None:
                     24,
                     80,
                     90,
-                    TRUE,
-                    'Automatic workspace reclamation is enabled by default.',
+                    FALSE,
+                    'Initial policy migrated from deployed retention behavior.',
                     'system',
                     TRUE
                 )
--- a/brain/platform/db/alembic/versions/0060_storage_policies.py
+++ b/brain/platform/db/alembic/versions/0060_storage_policies.py
@@ -124,6 +124,8 @@ def upgrade() -> None:
     active_count = op.get_bind().execute(
         sa.text("SELECT count(*) FROM storage_policies WHERE is_active = TRUE")
     ).scalar_one()
+    # This guarded seed only runs for fresh databases. Changing its defaults
+    # does not mutate deployments where this migration has already run.
     if active_count == 0:
         op.execute(
             sa.text(
@@ -144,8 +146,8 @@ def upgrade() -> None:
                     24,
                     80,
                     90,
-                    FALSE,
-                    'Initial policy migrated from deployed retention behavior.',
+                    TRUE,
+                    'Automatic workspace reclamation is enabled by default.',
                     'system',
                     TRUE
                 )
--- a/brain/platform/db/alembic/versions/0065_enable_automatic_reclamation.py
+++ b/brain/platform/db/alembic/versions/0065_enable_automatic_reclamation.py
@@ -1,4 +1,9 @@
-"""Enable automatic workspace reclamation for the untouched seed policy."""
+"""Enable automatic workspace reclamation for the untouched seed policy.
+
+Revision ID: 0065_enable_automatic_reclamation
+Revises: 0064_cycle_receipt_monitoring
+Create Date: 2026-09-01
+"""
 
 from __future__ import annotations
 
--- a/brain/platform/db/alembic/versions/0065_enable_automatic_reclamation.py
+++ b/brain/platform/db/alembic/versions/0065_enable_automatic_reclamation.py
@@ -1,0 +1,50 @@
+"""Enable automatic workspace reclamation for the untouched seed policy."""
+
+from __future__ import annotations
+
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = "0065_enable_automatic_reclamation"
+down_revision = "0064_cycle_receipt_monitoring"
+branch_labels = None
+depends_on = None
+
+_SEED_RATIONALE = "Initial policy migrated from deployed retention behavior."
+_ENABLED_RATIONALE = "Illospace issue #876 enabled automatic reclamation."
+
+
+def upgrade() -> None:
+    op.execute(
+        sa.text(
+            """
+            UPDATE storage_policies
+            SET automatic_reclamation_allowed = TRUE,
+                rationale = :enabled_rationale
+            WHERE is_active = TRUE
+              AND rationale = :seed_rationale
+            """
+        ).bindparams(
+            enabled_rationale=_ENABLED_RATIONALE,
+            seed_rationale=_SEED_RATIONALE,
+        )
+    )
+
+
+def downgrade() -> None:
+    op.execute(
+        sa.text(
+            """
+            UPDATE storage_policies
+            SET automatic_reclamation_allowed = FALSE,
+                rationale = :seed_rationale
+            WHERE is_active = TRUE
+              AND automatic_reclamation_allowed = TRUE
+              AND rationale = :enabled_rationale
+            """
+        ).bindparams(
+            enabled_rationale=_ENABLED_RATIONALE,
+            seed_rationale=_SEED_RATIONALE,
+        )
+    )
--- a/brain/platform/db/alembic/versions/0065_enable_automatic_reclamation.py
+++ b/brain/platform/db/alembic/versions/0065_enable_automatic_reclamation.py
@@ -16,6 +16,8 @@ down_revision = "0064_cycle_receipt_monitoring"
 branch_labels = None
 depends_on = None
 
+# Exact rationale literal written by the guarded seed in 0060_storage_policies;
+# it is the key that tells an untouched seed row from an operator edit.
 _SEED_RATIONALE = "Initial policy migrated from deployed retention behavior."
 _ENABLED_RATIONALE = "Illospace issue #876 enabled automatic reclamation."
 
--- a/brain/systems/cortex/project_context/materializer.py
+++ b/brain/systems/cortex/project_context/materializer.py
@@ -20,6 +20,7 @@ from brain.systems.cortex.project_context.identity import durable_project_id_fro
 from brain.systems.cortex.project_context.permissions import derive_project_permission_scope
 from brain.systems.cortex.project_context.project_root import is_synthetic_project_root_resource, project_key_from_context
 from brain.systems.cortex.project_context.resource_imports import (
+    ResourcePathAccess,
     backend_readable_resource_path,
     path_is_relative_to,
     runtime_workspace_path,
@@ -698,10 +699,15 @@ def _materialize_backend_readable_resource(
     project_context: dict[str, Any] | None = None,
 ) -> tuple[dict[str, str] | None, str | None, bool]:
     materialization = resource.get("materialization") if isinstance(resource.get("materialization"), dict) else {}
-    previous_source = should_use_existing_resource_path(_clean_text(materialization.get("source_path")), workspace_root)
+    previous_source = should_use_existing_resource_path(
+        _clean_text(materialization.get("source_path")),
+        workspace_root,
+        access=ResourcePathAccess.READ_ONLY,
+    )
     previous_draft = should_use_existing_resource_path(
         _clean_text(resource.get("path") or materialization.get("path")),
         workspace_root,
+        access=ResourcePathAccess.WRITE,
     )
     if previous_source and previous_draft and path_is_relative_to(previous_draft, workspace_root):
         existing_path = previous_draft
@@ -803,7 +809,11 @@ async def _materialize_resource(
             project_context=project_context,
         )
         return workspace, error
-    existing_path = should_use_existing_resource_path(_clean_text(resource.get("path")), workspace_root)
+    existing_path = should_use_existing_resource_path(
+        _clean_text(resource.get("path")),
+        workspace_root,
+        access=ResourcePathAccess.WRITE,
+    )
     if existing_path and path_is_relative_to(existing_path, workspace_root):
         return _workspace_entry_from_resource(resource, existing_path), None
 
--- a/brain/systems/cortex/project_context/resource_imports.py
+++ b/brain/systems/cortex/project_context/resource_imports.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 from collections.abc import Callable, Mapping
+from enum import Enum
 from pathlib import Path
 from typing import Any
 import os
@@ -14,6 +15,11 @@ from brain.systems.cortex.project_context.project_root import (
 
 
 ProjectSeedPredicate = Callable[[dict[str, Any]], bool]
+
+
+class ResourcePathAccess(str, Enum):
+    READ_ONLY = "read_only"
+    WRITE = "write"
 
 
 def clean_text(value: Any) -> str | None:
@@ -34,7 +40,13 @@ def is_managed_project_context_path(path: Path) -> bool:
     return ".illo-project-context" in path.expanduser().parts
 
 
-def should_use_existing_resource_path(existing_path: str | None, workspace_root: Path) -> Path | None:
+def should_use_existing_resource_path(
+    existing_path: str | None,
+    workspace_root: Path,
+    *,
+    access: ResourcePathAccess,
+) -> Path | None:
+    """Return an existing path without granting cross-workspace write access."""
     if not existing_path:
         return None
     path = Path(existing_path).expanduser()
@@ -42,7 +54,7 @@ def should_use_existing_resource_path(existing_path: str | None, workspace_root:
         return None
     if path_is_relative_to(path, workspace_root):
         return path
-    if is_managed_project_context_path(path):
+    if is_managed_project_context_path(path) and access is not ResourcePathAccess.READ_ONLY:
         return None
     return path
 
@@ -100,14 +112,22 @@ def backend_readable_resource_path(
 ) -> tuple[Path | None, bool]:
     for key in ("path", "local_path", "storage_path"):
         value = clean_text(resource.get(key))
-        existing_path = should_use_existing_resource_path(value, workspace_root)
+        existing_path = should_use_existing_resource_path(
+            value,
+            workspace_root,
+            access=ResourcePathAccess.READ_ONLY,
+        )
         if existing_path:
             return existing_path, True
 
     scoped_paths: list[Path] = []
     path_texts = [*_path_texts_from_uploaded_files(resource), *_path_texts_from_resource_scope(resource)]
     for value in path_texts:
-        existing_path = should_use_existing_resource_path(value, workspace_root)
+        existing_path = should_use_existing_resource_path(
+            value,
+            workspace_root,
+            access=ResourcePathAccess.READ_ONLY,
+        )
         if existing_path:
             scoped_paths.append(existing_path)
 
@@ -134,6 +154,7 @@ def _uploaded_file_import_candidates(
         source = should_use_existing_resource_path(
             clean_text(item.get("storage_path") or item.get("path") or item.get("local_path")),
             workspace_root,
+            access=ResourcePathAccess.READ_ONLY,
         )
         if not source or not source.is_file():
             continue
@@ -180,6 +201,7 @@ def project_native_import_candidates(
 
 __all__ = [
     "ProjectSeedPredicate",
+    "ResourcePathAccess",
     "backend_readable_resource_path",
     "clean_text",
     "path_is_relative_to",
--- a/brain/systems/host_capacity.py
+++ b/brain/systems/host_capacity.py
@@ -4,8 +4,8 @@ from __future__ import annotations
 import logging
 import os
 import shutil
-from collections.abc import Awaitable, Callable, Iterable
-from dataclasses import dataclass
+from collections.abc import Awaitable, Callable
+from dataclasses import dataclass, field
 from datetime import datetime
 from hashlib import sha256
 from pathlib import Path
@@ -58,6 +58,18 @@ class _CapacityAlert:
     alert_key: str
     alert_title: str
     error_text: Callable[[StoragePolicyValues], str]
+
+    def trigger_kind(self, mount: str) -> FailureGuardTriggerKind:
+        """Return this alert's scheduler-latch identity for one mount."""
+        scoped_key = f"{self.alert_key}:{mount}"
+        if len(scoped_key) <= 80:
+            return FailureGuardTriggerKind(scoped_key)
+        mount_digest = sha256(mount.encode("utf-8")).hexdigest()[:16]
+        digest_suffix = f":{mount_digest}"
+        mount_length = 80 - len(self.alert_key) - 1 - len(digest_suffix)
+        return FailureGuardTriggerKind(
+            f"{self.alert_key}:{mount[:mount_length]}{digest_suffix}"
+        )
 
 
 @dataclass(frozen=True, slots=True)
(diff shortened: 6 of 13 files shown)


redawear:

redawear:

Verified on `origin/main` by content probe, not by merge status: `HOST_CAPACITY_RECLAMATION_DISABLED_ALERT_KEY` at `brain/systems/host_capacity.py:49`, the migration `brain/platform/db/alembic/versions/0065_enable_automatic_reclamation.py`, and `should_use_existing_resource_path` at `brain/systems/cortex/project_context/materializer.py:27` are all present. Closing this original now that its content has shipped.

Issue #881, opened 2026-09-02
[triage] Every filed issue is auto-assigned to the workspace owner — the ownership playbook is not read on the filing path


redawear:
> *This was generated by AI during triage.*

## Summary

Every GitHub issue Illo files is assigned to the workspace owner (`redawear`) at filing time, regardless of who actually owns the work. Over the last six weeks this has put **18 tickets** on one person's board, **zero** of which a human ever assigned. The routing playbook that should decide the assignee exists and is not read on the filing path.

Illo diagnosed this itself when asked — see "Illo's own answer" below.

## Evidence

Measured 2026-09-02T16:15Z across `uwear-ai`. Every open issue assigned to `redawear` was assigned **0–21 seconds after its own creation, by the same actor that filed it**. Thirteen by `redawear` (an automation holding the owner's credentials), five by `illo-bot[bot]`:

| issue | created | assigned | by |
|---|---|---|---|
| uwear-ai/uwearaiapp#1322 | 2026-08-29T17:56:53Z | 17:56:54Z | `illo-bot[bot]` |
| uwear-ai/uwear-backend#1918 | 2026-08-28T14:22:07Z | same second | `illo-bot[bot]` |
| uwear-ai/uwearaiapp#1297 | 2026-08-22T13:21:18Z | 13:21:39Z | `illo-bot[bot]` |
| uwear-ai/uwearaiapp#1262 | 2026-08-17T12:53:04Z | same second | `illo-bot[bot]` |
| uwear-ai/uwear-backend#1125 | 2026-07-19T15:35:28Z | 15:35:42Z | `illo-bot[bot]` |

Two are demonstrably mis-routed under the org's own `uwear-engineering-triage` playbook (Enterprise Documentation, Domain 37, record 1155):

- **`uwear-backend#1918`** is MCP work. The playbook's specialization rule puts agent/LLM/MCP with `axel-havard`.
- **`uwear-backend#1125`** is a chantier whose **four remaining open sub-issues are all already assigned to `axel-havard`**.

## Illo's own answer, verbatim (run 21316)

> this is an Illo routing failure—not a GitHub default and not a choice you made.
>
> **GitHub does not default these tickets to you.** `create_github_issue` accepts an optional assignee list and passes it straight to GitHub. If Illo omits that list, the issue is created unassigned. […] Those near-instant assignment events mean the filing run explicitly supplied `redawear`.
>
[...]

PR #882, opened 2026-09-02
fix(triage): gate the filing path on a stated routing rationale, and tell an automation self-assignment from a human one

redawear:
> *This was generated by AI during an autonomous run.*

Two halves of [#881](https://github.com/Illospace/illospace/issues/881): stop the filing path
guessing an owner, and let a later read tell that guess apart from a human's.

## What was wrong

Every issue Illo files was assigned to the workspace owner at filing time — **18 tickets in six
weeks, none assigned by a human**, each assigned 0–21 seconds after its own creation by the same
actor that filed it. Two are mis-routed under the org's own triage playbook.

There is no hardcoded default anywhere. `create_github_issue` passes whatever assignee the
calling run supplies straight to GitHub. The run supplies the owner because **nothing makes it
justify the choice first**, and nothing downstream can tell that guess from a human decision.

## What this changes

**1. The filing path now gates on a stated rationale.**
`create_github_issue` takes `assignee_rationale`. Without one the assignees are dropped and the
issue is filed unassigned — it is still created, and the result carries `assignee_gate` and
`assignees_dropped` so the drop is visible rather than silent. The gate runs before the
token-candidate loop, so it is decided once, not once per candidate.

**2. Issue reads now carry `assignment_provenance`.**
One of `none`, `automation_at_filing`, `human`, `unknown`. An assignment counts as automation
only when *every* assigned event came from the issue's own creator within 60 seconds of creation
(a named constant covering the observed 0–21s range). The timeline is an extra API call, so it
is opt-in and defaults off; a disabled or failed fetch yields `unknown` and never blocks a read.
`unknown` and `human` stay distinct — a caller that cannot tell "I did not look" from "a human
did this" is the bug being fixed, one level up.

## What review caught

**During my own review**, the provenance classifier returned `automation_at_filing` for an issue
that **has** assignees but whose timeline carries no `assigned` event — asserting automation from
zero evidence. A truncated timeline page lands there, and the consuming routine filters on this
field, so a human-assigned ticket would have gone silently invisible. That case now returns
`unknown`, with a regression test for both shapes.

**A cross-family maintainability review** then returned BLOCKING with four findings. I verified
all four against the code before acting on them, and folded them in `6145e79a` — which removes
36 lines from the source:

1. `_github_timestamp` duplicated `_github_datetime`, already in the same module and strictly
   more capable. Deleted.
2. The **list-level** enrichment had no production caller and ran one fully-paginated timeline
   fetch per assigned issue, sequentially — up to 100 per page — with results held in a list
   coupled *by position* to the selection. Removed entirely. Provenance stays on the
   single-issue reader: no N+1, no positional coupling, and the shape a consumer checking one
   ticket actually needs. List reads now report `unknown`, the correct value for "I did not look".
3. The timeline handler caught bare `Exception`, turning real defects into ordinary `unknown`.
   Narrowed to `GitHubConnectorError` — the only error that path raises.
4. Not done, on purpose: extracting provenance into its own module with an enum contract. That
   module was oversized before this change, so that refactor belongs on its own, not bundled
   into a behaviour change.

Full review: `state/evidence/881-thermo-review-20260902T1905.md` in `uwear-routines`.

## Scope — and why #881 stays open after this merges

Deliberately **not** here: reading the live Domain 37 ownership record on the filing path, and
the precedence between an installed skill copy and that live record. The issue itself calls that
precedence *"undefined"*, so a worker would have been guessing at the contract.

This implements the conservative branch the issue's own acceptance states — *"or carries none"*,
*"or leaves them unassigned"* — plus the provenance signal. **No consumer reads
`assignment_provenance` yet**, so acceptance clause 3 is not finished. #881 needs the live-record
integration and a consumer before it can be closed. There is no closing keyword in this body on
purpose.

**One design decision is mine, not the issue's:** requiring a *stated* rationale rather than a
*verified* one. Verifying it needs the live playbook, which is out of scope above. The gate
therefore forces the caller to articulate routing before assigning; it does not judge whether the
routing is correct.

## Verification

| check | result |
|---|---|
| repo CI command verbatim, no path argument | **5052 passed, 11 skipped, 0 failed** (98.9s) — re-run AFTER the fix pass, not carried over |
| collected: `origin/main` → this branch | 5048 → 5063 (**+15**, and 5052+11 = 5063 — no coverage lost) |
| `meetbot/tests` included | yes — the 45 tests a `pytest tests/` shape drops |
| load average during the gate | **2.8–3.7**, far under the 20 that voids a gate |
| migrations | none — 0 alembic files touched |
| `frontend/**` files in the diff | **0**, so that lane is not selected |
| slice file overlap | zero — the two halves touch disjoint files |

## How to judge it without reading the diff

The behaviour is a tool contract, so the honest surface is the tool result:

1. File an issue through Illo with an assignee and **no** rationale → the issue is created
   **unassigned**, and the result carries `assignee_gate: "unjustified"` plus the dropped login.
2. File one with an assignee **and** a rationale → the assignee lands as before, with
   `assignee_gate: "justified"`.
3. Read any of the 18 auto-assigned tickets with provenance enabled → `automation_at_filing`.
   Read one you assigned by hand → `human`.

Refs #881.


redawear:

redawear:

Verified on `origin/main` by content probe, not by merge status: `GITHUB_AUTOMATION_ASSIGNMENT_WINDOW_SECONDS` and `_assignment_provenance` are present at `brain/systems/cortex/project_context/github.py:45,334`. Closing this original now that its content has shipped.

PR #883, opened 2026-09-03
fix(triage): make the assignment-provenance verdict reachable (#881 clause 3)

redawear:
Stacked on `illo-qa/881-assignment-routing` (PR #882). **Merge #882 first.**

#881 stays open after this merges — clauses 1 and 2 belong to #882.

## What this fixes

PR #882 builds the clause-3 detector and tests it well: `_assignment_provenance()` returns `none` / `unknown` / `human` / `automation_at_filing`, `_issue_payload()` surfaces it, and `async_get_issue(..., include_assignment_provenance=True)` fetches the timeline that feeds it.

Nothing could reach it. Measured on that branch, three independent probes:

1. `async_get_issue` had **zero production callers** — its definition plus three call sites in `tests/test_project_context_github.py`. (Every other `async_get_issue*` match in the repo is the unrelated `async_get_issue_closure_info`.)
2. The agent-callable `github` source-reader tool had **no `get_issue` action** and **no `issue_number` property**.
3. `include_assignment_provenance` appeared **nowhere** under `brain/systems/runs/`.

So the only issue read an agent could perform was `action="list_issues"` → `async_list_repo_issues`, which builds each payload with no timeline. `assignment_provenance` was `"unknown"` on every read any agent could actually make, and the playbook that would honour it still said "An explicit GitHub assignee always wins".

## What changed

- `definitions/github.py` — `get_issue` added to the source-reader action enum, plus an `issue_number` property. The action description spells out what each of the four verdicts means.
- `handlers/github.py` — a `get_issue` branch that validates `issue_number` before any token lookup and calls `async_get_issue(..., include_assignment_provenance=True)`. Provenance is **always on** for this action: a caller that has to know to ask for the safety signal will not ask. `issue_number` is threaded exactly the way `pull_number` already is.
- `uwear-engineering-triage/SKILL.md` — both ownership rules (the routing rule at ~389 and the Owner gate at ~486) now say *human* assignee, name `action="get_issue"` as the way to read the verdict, and say that `automation_at_filing` is Illo's earlier guess, so ownership is re-derived exactly as if the issue were unassigned.

## How to check it

This is a tool-surface and playbook change, so the reviewable surface is behaviour, not pixels:

1. Ask Illo to read `uwear-ai/uwear-backend#1918` with the `github` tool, `action="get_issue"`. The payload should carry `assignment_provenance: "automation_at_filing"` — that ticket was assigned by `illo-bot[bot]` in the same second it was filed.
2. Ask it to read a ticket a human assigned. The verdict should be `"human"`.
3. Ask it to triage #1918's ownership. Under the amended rule it should route by the playbook (MCP work → `axel-havard`) instead of deferring to the auto-assignment — which is #881's acceptance clause 2.
4. `action="get_issue"` with no `issue_number` returns `get_issue requires a positive issue_number` rather than raising.

## Verification — read this honestly

**The full-suite gate could not be run cleanly, so this PR does not claim one.**

The targeted file passes: `tests/test_github_source_tool.py` → **46 passed** (run twice, 1.70s and 8.28s). That covers the new action, the `issue_number` guard, and the schema contract.

The repo CI command verbatim (no path argument, so `meetbot/tests` is included) was started at load 12.6 and never finished. Two things spoiled it: the machine reached **load 54**, and the Codex worker recovered from its stream-reconnect and launched a second copy of the same suite 51 seconds after mine. Two concurrent suites under that load is not evidence in either direction, so no green is quoted here.

No `frontend/**` file and no alembic revision is in this layer, so neither of those lanes is selected — the Python lane is the only one this diff triggers.

## Notes for the reviewer

- The Codex worker that wrote the production change died in stream-reconnect backoff before it could add tests or commit. The orchestrator reviewed the diff, wrote the four tests in `tests/test_github_source_tool.py`, and committed. Flagging it so the authorship is not a surprise.
- `_read_with_token` has exactly one caller, which passes the new keyword — checked, because a new required keyword-only argument would otherwise break it silently.

Closes nothing on its own.

### DIFF
--- a/brain/systems/runs/tool_catalog/definitions/github.py
+++ b/brain/systems/runs/tool_catalog/definitions/github.py
@@ -20,6 +20,7 @@ GITHUB_TOOLS = [
                     "type": "string",
                     "enum": [
                         "get_repo",
+                        "get_issue",
                         "list_issues",
                         "list_pull_requests",
                         "get_pull_request",
@@ -31,7 +32,11 @@ GITHUB_TOOLS = [
                     ],
                     "default": "list_issues",
                     "description": (
-                        "Which GitHub source data to read. get_pull_request includes mergeability, CI "
+                        "Which GitHub source data to read. get_issue reads one issue by exact number and "
+                        "returns assignment_provenance: none means no assignee, unknown means provenance could "
+                        "not be determined, human means a human assignment, and automation_at_filing means the "
+                        "filing actor assigned it within seconds of creation, so it is not a human assignment. "
+                        "get_pull_request includes mergeability, CI "
                         "check runs, and combined status; pull_request_checks reads CI for a head SHA; "
                         "get_counts returns exact issue and PR counts for the requested state. get_file, "
                         "list_tree, and grep require ref; grep is literal rather than regular-expression search."
@@ -67,6 +72,11 @@ GITHUB_TOOLS = [
                     "type": "integer",
                     "minimum": 1,
                     "description": "Pull request number required by get_pull_request.",
+                },
+                "issue_number": {
+                    "type": "integer",
+                    "minimum": 1,
+                    "description": "Issue number required by get_issue.",
                 },
                 "number": {
                     "type": "integer",
--- a/brain/systems/runs/tool_catalog/handlers/github.py
+++ b/brain/systems/runs/tool_catalog/handlers/github.py
@@ -21,6 +21,7 @@ from brain.systems.cortex.project_context.github import (
     async_add_repo_sub_issue,
     async_create_repo_issue,
     async_grep_repo,
+    async_get_issue,
     async_get_repo_issue_parent,
     async_get_issue_closure_info,
     async_get_pull_request_deploy_info,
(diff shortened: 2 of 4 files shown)


redawear:

redawear:
⚠️ **Do not merge this branch on its own — it breaks `main`.**

It edits `brain/systems/skills/builtin_skill_bundles/uwear-engineering-triage/SKILL.md` without updating the mirror fingerprint that `docs/329-live-delta.md` pins, so `tests/test_builtin_skill_suite.py::test_uwear_triage_memory_delta_fingerprints_exact_bundle_mirrors` fails on the `v9` mirror.

Deterministic, not contention: it reproduces in 0.3s on this branch and passes on `origin/main`. This PR's own body says its full-suite gate never finished — this is what it would have caught.

Fixed in #884 as `894eda85` (characters 32262 → 32770, bytes 32358 → 32866, SHA-256 1344faa5… → 43ee7178…), the same three-line shape every SKILL.md edit has used since 2026-07-07.

redawear:

Verified on `origin/main` by content probe, not by merge status: `get_issue requires a positive issue_number` at `brain/systems/runs/tool_catalog/handlers/github.py:517`, `include_assignment_provenance=True` at line 384, and `automation_at_filing` in the triage `SKILL.md` are all present. Closing this original now that its content has shipped.

PR #893, opened 2026-09-07
fix(alerts, github-tool): claim alert filing atomically, and stop an omitted update field meaning "destroy"

### LINKED ISSUE #892: [alerts] Concurrent signature dedup is non-atomic, and required update args can clobber issue bodies

## Summary

Four simultaneous Rollbar messages for one production failure entered the monitored Slack `#alerts` lane at 2026-09-06 17:49 UTC. They shared the exact tracked signature:

`207d7d2ef53b3e725ba969c300affc552f2dd2ca4c3eaeb8ba01bf3e2e4a5c1f`

Three independent runs each completed a read-before-write duplicate search and then opened a separate real GitHub issue before any canonical issue became visible to the other runs:

- `uwear-ai/uwear-backend#2021`, created `17:51:36Z` for Rollbar `Uwear-API#2411`
- `uwear-ai/uwear-backend#2022`, created `17:52:12Z` for `Uwear-API#2413`
- `uwear-ai/uwear-backend#2023`, created `17:52:20Z` for `Uwear-API#2414`

The fourth message (`#2412`) was eventually added as an occurrence of #2021 by continuation run 22416, but only after #2021/#2022/#2023 already existed. Domain 1 likewise acquired three tracker records: 6005, 6007, and 6008.

The same incident also reproduced destructive issue-update behavior during post-create enrichment. Domain record 6007's audit note states that assigning #2022 accidentally cleared its body and labels and that a reconstructed body/labels had to be restored. Exact GitHub reads during the repair window observed #2022 with `body=null`, and a separate exact read of #2023 around 17:57 observed only the 43-character AI provenance banner. Current readback is repaired (#2022: 2,696 body chars; #2023: 2,581 chars), so this was a transient integrity failure, not permanent current loss.

## Why the duplicate check races

The monitored intake already computes and persists deterministic provider-alert identity before starting triage:

- `brain/systems/slack/monitored_intakes.py:479-523` calls `handle_provider_alert_ingest_durable` and puts `tracked_signature` into the run payload.
- `brain/platform/db/models/provider_alert.py:22-70` defines one `provider_alert_ledger` row per `(org_id, channel_id, signature)`.
- `provider_alert_occurrences` is idempotent only by `(org_id, channel_id, slack_message_ts)` and indexes, but does not uniquely claim ticket creation by signature.
- `brain/systems/slack/provider_alert_surge.py:385-430` records each message/signature occurrence. It does not reserve a canonical ticket or gate later filing.
- `brain/systems/slack/channel_monitor_rendering.py:110-117` and `uwear-engineering-triage/references/creating-work-items.md:75-87` instruct the model to search and not refile an exact signature, but that is a non-atomic read-before-write policy.
- `create_github_issue` has no signature/idempotency/claim argument. `brain/systems/runs/tool_catalog/handlers/github.py:651-735` goes directly from validation/token selection to `async_create_repo_issue`.

Concurrent runs can therefore all observe “no match” and all write. The provider ledger has the right stable key but does not own or lease the side effect.

## Why enrichment can erase body/labels

The checked-in source contract says only `repo` and `issue_number` are required and that omitted fields stay unchanged:

- `brain/systems/runs/tool_catalog/definitions/github.py:260-334`

But the live agent-facing tool schema currently presents every update field as mandatory (`assignees_add`, `assignees_remove`, `labels_add`, `labels_remove`, `labels_set`, `state`, `title`, `body`, and `token_secret_key`). A caller performing an assignee-only update is therefore forced to synthesize values for fields it does not intend to touch.

The handler then gives those synthetic values destructive meaning:

- `labels_set=[]` means replace all labels with none.
- `body=""` becomes `clean_body=""` and is sent as a replacement, clearing the body.
- `brain/systems/runs/tool_catalog/handlers/github.py:1038-1165` applies every non-`None` field independently.

This is the same failure family as closed #870/#871, but it is reproducible on the current live tool surface after that fix: the source JSON schema is optional while the exported runtime schema given to the model is required. The current incident proves the mismatch still reaches real issues.

## Expected behavior

1. Before launching or admitting a monitored-alert filing run, atomically resolve `(org_id, service, subsystem, tracked_signature)` to one canonical filing claim/ticket reference.
2. Exactly one run may own the initial GitHub create side effect. Other simultaneous occurrences wait for or reuse the canonical issue, then append occurrence evidence.
3. Persist claim state and resulting `repo`, `issue_number`, and URL durably; support lease expiry/retry when the owner run fails before or after GitHub creation.
4. Do not rely on GitHub/Domain search visibility for concurrency correctness. Searches remain useful for historical dedup, but not as the lock.
5. Export `update_github_issue` fields as truly optional end to end. An assignee-only call must serialize only the assignee delta.
6. Reject ambiguous synthetic defaults at the tool boundary. Omission must not become `body=""` or `labels_set=[]`.
7. For any requested title/body/label replacement, read back and compare the requested postcondition; report partial failure without silently continuing.

## Acceptance tests

- Admit four concurrent Rollbar messages with different Slack timestamps/external item ids but the same tracked signature. Assert one GitHub create call, one canonical ticket reference, four occurrences, and no duplicate tracker records.
- Crash the filing owner before GitHub creation; verify another run can acquire an expired lease and create exactly once.
- Simulate a successful GitHub create followed by local finalization failure; retry must discover/reconcile the created issue rather than create another.
- Run an assignee-only `update_github_issue` through the actual exported model tool schema. Assert title, body, labels, state, and existing assignees are byte-for-byte unchanged except for the requested assignee delta.
- Repeat for label-only, state-only, and comment-only enrichment.
- Add a contract test comparing the checked-in JSON schema to the runtime/provider-exported schema so optional fields cannot silently become required again.

## Incident references

- Parent alert run: 22411
- Dedup worker: 22415 (reported no exact match while complete authenticated search was unavailable)
- Joined continuation: 22416 (found #2021/#2022/#2023 and Domain 6005/6007/6008)
- Filing runs: 22407 (#2021), 22409 (#2022), 22410 (#2023)
- Enrichment/repair runs: 22417 (#2023), 22418 (#2022)
- Slack origin thread: `slack:T049MELA59D:C082SUKQKJL:1788716941.540069`
- Existing related issue: #870, closed by #871; this incident demonstrates the runtime export still violates that intended contract.

No Uwear issue, Domain record, or Slack message was modified while preparing this report.

redawear:
Closes #892.

**This PR changed shape after it was opened.** It originally carried only the
second of #892's two defects, and its stacked child #894 carried the first. #894
merged into this branch on 2026-09-07 at 15:16Z, so this single PR now carries
**both** halves and closes the issue on merge. The train is gone — there is one
click here, not two.

| commit | half | what it fixes |
|---|---|---|
| `ceb89e0e` | Part B | an omitted update field could mean "destroy" |
| `cfabf6ab` | Part A | the provider-alert filing claim was not atomic |

## Part A — three runs each filed their own issue for one alert

Four simultaneous Rollbar messages for one production failure shared the exact
tracked signature. Three independent runs each completed a read-before-write
duplicate search, each saw "no match", and each opened a real GitHub issue:
`uwear-ai/uwear-backend#2021`, `#2022`, `#2023`. Domain 1 gained three tracker
records for one fault.

The provider ledger already had the right stable key and did not own the side
effect. A search is not a lock.

This adds an atomic filing claim on `(org_id, service, subsystem,
tracked_signature)` — new table via migration `0066_provider_alert_filing_claims`,
logic in `brain/systems/slack/provider_alert_filing.py`. Exactly one run owns the
GitHub create. Others wait for the canonical issue and append occurrence evidence.
The claim is durable and leased, so a run that dies before or after the create
does not strand the alert and does not file twice.

`main` already carried a same-signature **admission** fence from #620. That fence
guards admission; the filing side effect happened later and unguarded, which is
why it did not stop this incident.

## Part B — an omitted update field could mean "destroy"

On 2026-09-06 an assignee-only enrichment call **cleared the body and labels** of
`uwear-ai/uwear-backend#2022`, and `#2023` was observed carrying only its
43-character provenance banner. Both were repaired by hand afterwards.

`_handle_update_github_issue` applied every non-`None` field independently, so the
values a caller synthesizes when it believes a field is mandatory both carried
destructive meaning: `body=""` replaced the body with nothing, and `labels_set=[]`
replaced every label with none.

Destruction is now unreachable by accident and still reachable on purpose:

- `body=""` and `labels_set=[]` are **rejected** with a 422 that names the alternative.
- Clearing is explicit: `clear_body: true`, `clear_labels: true`.
- Each flag refuses to be combined with the field it replaces, so there is one way
  to express each intent.

### The issue's other hypothesis was wrong, and this PR proves it

#892 says *"the live agent-facing tool schema currently presents every update field
as mandatory"*. **They are already optional.** Measured through both transports and
both tool sets, the exported schema carries `required: ["repo", "issue_number"]` —
identical to the checked-in definition. The two functions that do force every
property into `required` (`brain/systems/memory/truth_maintenance.py:915`,
`brain/systems/sessions/harvest_extraction.py:520`) serve memory structured outputs
and never receive the GitHub tool schema.

`tests/test_github_issue_export_contract.py` pins this so the hypothesis cannot come
back unmeasured.

## Why this closes #892

All seven of the issue's expected behaviors are covered, and all six of its
acceptance tests exist:

| # | expected behavior | where |
|---|---|---|
| 1–3 | atomic claim, single create owner, durable claim + lease expiry | Part A |
| 4 | search is not the lock | Part A |
| 5–6 | update fields optional end to end; no synthetic destructive defaults | Part B |
| 7 | read back the requested postcondition, report partial failure | already on `main` (`brain/systems/cortex/project_context/github.py:2683`) |

Acceptance tests, by name in `tests/test_provider_alert_filing.py`:
`test_four_messages_force_claim_conflict_and_share_one_ticket` (four concurrent
messages, one create), `test_dead_owner_before_create_is_replaced_after_expiry`
(crash before create, expired lease, exactly one create),
`test_create_survives_failed_finalization_and_reconciles` (create then local
finalization failure reconciles instead of re-creating). The update-path scenarios
are in `tests/test_update_github_issue_tool.py`, and the schema contract test is
`tests/test_github_issue_export_contract.py`.

## Verification

Repo CI command verbatim from `.github/workflows/brain-ci.yml`, no path argument so
`meetbot/tests` is included:

```
python3 -m pytest -m "not requires_db and not requires_browser and not live_provider" -q
```

Measured on this PR's head `cfabf6ab` — the exact commit you would merge:

| | this PR (`cfabf6ab`) | clean `origin/main` (`0fa3e0e6`) |
|---|---|---|
| passed | **5176** | 5130 |
| failed | 2 | **1** |
| collected | 5178 | 5131 |
| wall clock | 110.5s | 108.0s |

**The two reds are machine artifacts, not this diff — proven, not assumed.**
`tests/test_safe_deploy.py::test_worker_restart_policy_suspension_is_restored_when_a_deploy_is_interrupted[HUP-1-129]`
fails **identically on unmodified `origin/main`**, which is the baseline column
above. `tests/test_self_update_daemon.py::test_keeper_stops_when_controller_stops[2]`
failed here and not on the baseline; both pass in isolation on this same tree
(`7 passed in 1.23s`). Both tests exercise the deploy and
self-update daemons — neither touches alert filing or the GitHub tool surface this PR
changes.

To be precise about what that evidence does and does not establish: it shows these two
tests are **nondeterministic**, and it rules out this diff as the cause. It does not
establish *why* they flake. Load is the usual suspect here and was 4.4–5.3 throughout,
but this run did not isolate it, so treat the mechanism as unproven and the "not this
diff" conclusion as the claim that matters.

The +47 collected delta reconciles against the new tests: `test_provider_alert_filing.py`
(new, 13 tests), `test_github_issue_export_contract.py` (new), and the additions to
`test_update_github_issue_tool.py`.

`alembic heads` prints exactly one: `0066_provider_alert_filing_claims`. The diff
touches no `frontend/**` file, so that CI lane does not apply.

## How to judge it without reading the diff

Part B is a tool contract, so exercise it:

1. Call `update_github_issue` with only `repo`, `issue_number` and `assignees_add`.
   Title, body, labels and state must come back byte-for-byte unchanged.
2. Call it with `body: ""`. It must fail with a 422 telling you to omit `body` or use
   `clear_body`, and the issue must be untouched.
3. Call it with `clear_body: true` and no `body`. The body must actually clear — the
   escape hatch has to work, or this trades one bug for another.
4. Call it with both `clear_labels: true` and `labels_set: [...]`. It must refuse
   rather than pick one.

Part A needs a real concurrent alert to exercise by hand. The closest honest check is
the first test named above, which forces the claim conflict directly.

## Known follow-up, already filed

A code-quality pass on Part A returned one **BLOCKING** maintainability finding that
was deliberately **not** folded here, so this PR's verified gate stays attached to an
unchanged commit: `_handle_create_github_issue` now runs two error policies
interleaved in one token-fallback loop, branching on `filing_identity` in three
places. Filed as #895 rather than left inside a merged PR body.

Part B's review returned APPROVE_WITH_NOTES. Its one note is test organization only:
`tests/test_update_github_issue_tool.py:365` exercises comment preservation and
belongs in `tests/test_add_github_issue_comment_tool.py`. No shipped behavior.

Three NOTE-level findings on Part A stay recorded and unfixed: `claimed_at` doubling
as lease age and generation identity, the undocumented 60s lease against the ~15s wait
budget, and the body-marker serialization being an implicit external contract.

**Roles, stated plainly:** Codex authored both diffs, and a second Codex pass ran the
maintainability rubric — those two are same-family, so the rubric is not an
independent check of the author. The cross-family element is the Claude director
review: the stated cause was verified against `origin/main` before dispatch, the diffs
were read directly, the acceptance coverage above was probed with positive and
negative controls, and every gate number in this section was measured by the director,
not carried from a worker.

---

## Merge order and merge method (added 2026-09-09)

The "the train is gone" note above is about #894 and #892's two halves. It is not
the whole picture: **PR #896 is stacked on this branch** and is a second click,
for a different issue.

Merge this one with **"Create a merge commit"** or **"Rebase and merge"** — **not
"Squash and merge"** — then merge #896. Two independent reasons, both measured on
2026-09-09:

1. This PR **adds** 4 files, and #896's branch adds those same 4 plus one more.
   A squash rewrites this branch's history, so #896 would then land add/add
   conflicts on all four.
2. #896's body carries no linking keyword of its own; its link to its issue lives
   only in commit `cd2befb1`'s message, which a squash rewrites.

Gate re-run on #896's tip (`596e5cef`, tree `010dfd09`) on 2026-09-09: 5178
passed, 11 skipped, 0 failed; `alembic heads` prints exactly one.

### DIFF
--- a/brain/platform/db/alembic/versions/0066_provider_alert_filing_claims.py
+++ b/brain/platform/db/alembic/versions/0066_provider_alert_filing_claims.py
@@ -1,0 +1,47 @@
+"""Reserve one canonical GitHub filing per provider-alert signature.
+
+Revision ID: 0066_provider_alert_filing_claims
+Revises: 0065_enable_automatic_reclamation
+Create Date: 2026-09-07
+"""
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+revision = "0066_provider_alert_filing_claims"
+down_revision = "0065_enable_automatic_reclamation"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # The public baseline already materializes current models on fresh installs.
+    if sa.inspect(op.get_bind()).has_table("provider_alert_filing_claims"):
+        return
+    op.create_table(
+        "provider_alert_filing_claims",
+        sa.Column("id", sa.Integer(), autoincrement=True, primary_key=True),
+        sa.Column("org_id", postgresql.UUID(as_uuid=False).with_variant(sa.String(), "sqlite"),
+                  sa.ForeignKey("orgs.id", ondelete="CASCADE"), nullable=False),
+        sa.Column("service", sa.String(120), nullable=False),
+        sa.Column("subsystem", sa.String(120), nullable=False),
+        sa.Column("tracked_signature", sa.String(64), nullable=False),
+        sa.Column("repo", sa.String(255), nullable=False),
+        sa.Column("state", sa.String(20), server_default=sa.text("'pending'"), nullable=False),
+        sa.Column("claimed_at", sa.DateTime(timezone=True), nullable=True),
+        sa.Column("issue_number", sa.Integer(), nullable=True),
+        sa.Column("issue_url", sa.Text(), nullable=True),
+        sa.Column("created_at", sa.DateTime(timezone=True), server_default=sa.text("now()"), nullable=False),
+        sa.Column("updated_at", sa.DateTime(timezone=True), server_default=sa.text("now()"), nullable=False),
+        sa.UniqueConstraint("org_id", "service", "subsystem", "tracked_signature",
+                            name="uq_provider_alert_filing_signature"),
+    )
+    op.create_index("ix_provider_alert_filing_pending", "provider_alert_filing_claims", ["state", "claimed_at"])
+
+
+def downgrade() -> None:
+    if not sa.inspect(op.get_bind()).has_table("provider_alert_filing_claims"):
+        return
+    op.drop_index("ix_provider_alert_filing_pending", table_name="provider_alert_filing_claims")
+    op.drop_table("provider_alert_filing_claims")
--- a/brain/platform/db/models/provider_alert.py
+++ b/brain/platform/db/models/provider_alert.py
@@ -14,10 +14,39 @@ from brain.platform.db.base import Base, TimestampMixin
 UUIDString = UUID(as_uuid=False).with_variant(String, "sqlite")
 
 __all__ = [
+    "ProviderAlertFilingClaim",
     "ProviderAlertLedger",
     "ProviderAlertOccurrence",
     "ProviderAlertSurge",
 ]
+
+
+class ProviderAlertFilingClaim(Base, TimestampMixin):
+    """Canonical GitHub filing and expiring owner for a tracked signature."""
+
+    __tablename__ = "provider_alert_filing_claims"
+    __table_args__ = (
+        UniqueConstraint(
+            "org_id", "service", "subsystem", "tracked_signature",
+            name="uq_provider_alert_filing_signature",
+        ),
+        Index("ix_provider_alert_filing_pending", "state", "claimed_at"),
+    )
+
+    id: Mapped[int] = mapped_column(Integer, primary_key=True, autoincrement=True)
+    org_id: Mapped[str] = mapped_column(
+        UUIDString, ForeignKey("orgs.id", ondelete="CASCADE"), nullable=False,
+    )
+    service: Mapped[str] = mapped_column(String(120), nullable=False)
+    subsystem: Mapped[str] = mapped_column(String(120), nullable=False)
+    tracked_signature: Mapped[str] = mapped_column(String(64), nullable=False)
+    repo: Mapped[str] = mapped_column(String(255), nullable=False)
+    state: Mapped[str] = mapped_column(
+        String(20), nullable=False, default="pending", server_default=text("'pending'"),
+    )
+    claimed_at: Mapped[datetime | None] = mapped_column(DateTime(timezone=True), nullable=True)
+    issue_number: Mapped[int | None] = mapped_column(Integer, nullable=True)
+    issue_url: Mapped[str | None] = mapped_column(Text, nullable=True)
 
 
 class ProviderAlertLedger(Base, TimestampMixin):
--- a/brain/systems/runs/tool_catalog/definitions/github.py
+++ b/brain/systems/runs/tool_catalog/definitions/github.py
@@ -269,7 +269,8 @@ GITHUB_TOOLS = [
             "as total success. This uses the same project-bound GitHub App write identity as issue "
             "creation; the resulting GitHub issues webhook keeps configured mirrored ticket records "
             "in sync. labels_set replaces every label and is mutually exclusive with labels_add and "
-            "labels_remove."
+            "labels_remove. Empty body or labels_set values are rejected; omit them to preserve "
+            "existing content. Use clear_body or clear_labels only for an intentional clear."
         ),
         "input_schema": {
             "type": "object",
@@ -307,8 +308,16 @@ GITHUB_TOOLS = [
                     "type": "array",
                     "items": {"type": "string"},
                     "description": (
-                        "Replace all issue labels with this exact list; [] clears every label. "
+                        "Replace all issue labels with this non-empty list. To clear all labels, "
+                        "omit labels_set and use clear_labels: true. "
                         "Do not combine with labels_add or labels_remove."
+                    ),
+                },
+                "clear_labels": {
+                    "type": "boolean",
+                    "description": (
+                        "Explicitly remove every issue label. Omit labels_set, labels_add and "
+                        "labels_remove when true. Omit this flag to preserve labels."
                     ),
                 },
                 "state": {
@@ -322,7 +331,14 @@ GITHUB_TOOLS = [
                 },
                 "body": {
                     "type": "string",
-                    "description": "Replacement Markdown body; an empty string clears the body.",
+                    "description": (
+                        "Non-empty replacement Markdown body. To clear the body, omit body and "
+                        "use clear_body: true."
+                    ),
+                },
+                "clear_body": {
+                    "type": "boolean",
+                    "description": "Explicitly clear the issue body. Omit body when true.",
                 },
                 "token_secret_key": {
                     "type": "string",
--- a/brain/systems/runs/tool_catalog/definitions/github.py
+++ b/brain/systems/runs/tool_catalog/definitions/github.py
@@ -202,6 +202,20 @@ GITHUB_TOOLS = [
                         "to deliver the created artifact back to a matching open request."
                     ),
                 },
+                "provider_alert": {
+                    "type": "object",
+                    "description": (
+                        "Optional deterministic provider-alert identity. Atomically reserves the initial "
+                        "filing or returns the canonical issue and appends occurrence evidence from body. "
+                        "If filing_pending is returned, retry with the same identity."
+                    ),
+                    "properties": {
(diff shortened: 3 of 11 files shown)


redawear:
> *This was generated by AI during an autonomous R3 run.*

**Merge this one with a MERGE COMMIT, not a squash.** PR #896 is stacked on this branch, so the method you pick decides whether #896 survives the merge.

- **Merge commit** — keeps ancestry. #896's net diff against `main` stays its own 2 files, and it can then be retargeted to `main` cleanly. Measured on this repo before (#836/#837, 2026-08-17).
- **Squash** — puts this content on `main` with no shared ancestry. #896 then hits add/add conflicts on every file this PR **creates**: `brain/systems/slack/provider_alert_filing.py`, `brain/systems/runs/tool_catalog/handlers/github_issue_create.py`, migration `0066_provider_alert_filing_claims`, and their tests. Recoverable, but it costs a repair pass.

All three merge methods are enabled on the repo, so nothing stops the squash by itself.

**You are not blocked from merging.** This PR reads `mergeStateStatus: BLOCKED` only because it is still a draft. The "Protect main" ruleset lists `redawear` as a bypass actor with `bypass_mode: pull_request` (re-measured 2026-09-07), so the normal merge button works once you un-draft. It requires 0 approvals and has no required status checks.

**Gate, re-run this session rather than carried:** the merged tree `596e5cef` (#893 + #896 — #896's tip contains this PR's head `cfabf6ab`) gives **5178 passed, 0 failed, 11 skipped**, 103.4s, load 3.2, serial. Repo CI command verbatim from `.github/workflows/brain-ci.yml:116-120`, no path argument so `meetbot/tests` runs. `alembic heads` prints exactly one. Frontend lane not triggered — 0 of the 12 changed files are under `frontend/`. This is the second independent sample of that tree at the same figure.

After this merges, #896 needs `gh pr edit 896 --base main` **and** a body edit — GitHub does not re-parse the closing keyword on retarget alone, so #895 stays open until it does.

PR #894, opened 2026-09-07
fix(alerts): make the provider-alert filing claim atomic

redawear:
> Part A of #892. **Stacked on #893** (Part B) — merge that first, or merge this train in one click.

## What went wrong

On 2026-09-06 at 17:49 UTC, four Rollbar messages for **one** production failure arrived in the monitored Slack `#alerts` lane sharing the exact tracked signature `207d7d2e…a5c1f`.

Three independent runs each completed a read-before-write duplicate search, each saw "no match", and each filed a **separate real GitHub issue** — `uwear-ai/uwear-backend#2021`, `#2022`, `#2023` — plus three tracker records (Domain 1: 6005, 6007, 6008).

The duplicate check was a **policy the model was asked to follow**, not a lock. The provider-alert ledger already held the right stable key, but nothing made it own the side effect.

## What this changes

A `provider_alert_filing_claims` row now owns the filing.

- The canonical key `(org_id, service, subsystem, tracked_signature)` is inserted with `on_conflict_do_nothing`, then its expiring claim column is taken by **compare-and-set**. Exactly one run can create the issue.
- The claim moves to `creating` and **commits before** the GitHub request. So a crash *after* the request leaves a durable instruction to reconcile rather than to file again.
- Recovery matches a hashed marker in the issue body through the repository issues endpoint, **including closed issues** — never the search index, which is what failed here.
- Concurrent runs reuse the canonical issue and append their occurrence as a comment.
- The claim **expires**, so an owner that dies before or after the create does not wedge the signature forever.

`create_github_issue`'s new `provider_alert` argument is **optional**. Omit it and behavior is exactly as today — deliberately, because #893 exists precisely because callers were once forced to synthesize values for fields they did not intend to touch.

## How to see it work

This is Illo, so the real gate is using it. After merge, the next time two alerts share a signature in `#alerts`:

1. Confirm **one** issue is filed, not three.
2. The later occurrences appear as **comments on that one issue**.
3. Domain 1 gains **one** tracker record for the signature.
4. The issue body carries an `<!-- illo-provider-alert-filing:… -->` marker — that is the recovery key; leave it in place while filing is unresolved.

## Verification

| | |
|---|---|
| suite | **5178 passed**, 0 failed, 11 skipped, 90 deselected |
| commit | `027a6f99` — the head of this PR |
| command | repo CI verbatim, no path argument, so `meetbot/tests` is included |
| wall clock | 94.22s, serial, load 3.34, no Codex worker live |
| alembic | one head — `0066_provider_alert_filing_claims` |
| new tests | 17 in `tests/test_provider_alert_filing.py`, covering the ticket's three acceptance cases |

The parent branch measures 5161, and 5161 + 17 = 5178 — the delta is reconciled, not assumed.

## Review — and an honest label

Codex authored this diff, and a second Codex pass ran the thermo-nuclear maintainability rubric (19 files read): **BLOCKING, 2 blockers + 3 notes**. Those two passes are the **same family**, so the rubric run does not independently check the author. The cross-family element is the Claude director review: pre-dispatch scan, reading the diff, re-running the gate outside the sandbox, reconciling the delta, and triaging the blockers.

**Blocker 2 is fixed in `027a6f99`.** The injected `create_issue` callback plus `**kwargs` hid the title/body/token the module actually requires, while reconciliation and commenting already called the connector directly. Explicit typed parameters now; no callback.

**Blocker 1 is deliberately NOT fixed here**, and it is the one thing worth your judgement:

> `handlers/github.py` now selects two error policies inside one token-fallback loop — ordinary creation re-raises unexpected exceptions, provider-alert filing converts them to retryable pending results.

The reviewer's fix is to split the plain-create and provider-alert adapters and extract the shared token loop. That restructures a handler which was **already over 1000 lines before this PR**, so folding it here would widen the blast radius well past this ticket and detach the verified 5178 from a much larger change. It belongs in its own pass.

Three NOTE-level findings are also left for later, recorded so they are not lost: `claimed_at` doubles as lease age *and* generation identity (an opaque token would separate them); the 60s lease versus the ~15s wait budget is undocumented; and the body-marker serialization is an implicit external contract that deserves a versioned codec.

## Scope

#892 stays open after this merges — it carries two defects, and #893 is the other one.

### DIFF
--- a/brain/platform/db/alembic/versions/0066_provider_alert_filing_claims.py
+++ b/brain/platform/db/alembic/versions/0066_provider_alert_filing_claims.py
@@ -1,0 +1,47 @@
+"""Reserve one canonical GitHub filing per provider-alert signature.
+
+Revision ID: 0066_provider_alert_filing_claims
+Revises: 0065_enable_automatic_reclamation
+Create Date: 2026-09-07
+"""
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+revision = "0066_provider_alert_filing_claims"
+down_revision = "0065_enable_automatic_reclamation"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # The public baseline already materializes current models on fresh installs.
+    if sa.inspect(op.get_bind()).has_table("provider_alert_filing_claims"):
+        return
+    op.create_table(
+        "provider_alert_filing_claims",
+        sa.Column("id", sa.Integer(), autoincrement=True, primary_key=True),
+        sa.Column("org_id", postgresql.UUID(as_uuid=False).with_variant(sa.String(), "sqlite"),
+                  sa.ForeignKey("orgs.id", ondelete="CASCADE"), nullable=False),
+        sa.Column("service", sa.String(120), nullable=False),
+        sa.Column("subsystem", sa.String(120), nullable=False),
+        sa.Column("tracked_signature", sa.String(64), nullable=False),
+        sa.Column("repo", sa.String(255), nullable=False),
+        sa.Column("state", sa.String(20), server_default=sa.text("'pending'"), nullable=False),
+        sa.Column("claimed_at", sa.DateTime(timezone=True), nullable=True),
+        sa.Column("issue_number", sa.Integer(), nullable=True),
+        sa.Column("issue_url", sa.Text(), nullable=True),
+        sa.Column("created_at", sa.DateTime(timezone=True), server_default=sa.text("now()"), nullable=False),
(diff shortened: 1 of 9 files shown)


redawear:
> *This was generated by AI during an autonomous R3 run.*

The **BLOCKING** review finding this PR names but does not fold — two error policies interleaved inside the one token-fallback loop in `_handle_create_github_issue` — is now filed as #895, so it survives the merge instead of living only in this body.

The three NOTE-level findings stay recorded here and are listed as out of scope on that ticket.

Gate re-measured on this PR's exact head `027a6f99` by a later run: **5178 passed, 0 failed, 11 skipped**, 100.93s, serial, load 2.7, no worker live. One alembic head (`0066_provider_alert_filing_claims`).

Issue #905, opened 2026-09-21
[tracker] A Domain-1 write at 2026-09-21 12:02Z made 5 duplicate records with an existing external_id


redawear:
> *This was generated by AI during triage.*

## What happens

A Domain-1 (GitHub Ticket Tracker) write at **2026-09-21 12:02Z** made new records that copy existing ones. Each copy has the same `object_key`, `data.repo`, `data.number` and `external_id` title as a record that already exists.

| new record | copies | external id | status (both) |
|---|---|---|---|
| 6857 | 6793 | `github:uwear-ai/uwearaiapp:pr:1485` | Done |
| 6858 | 6794 | `github:uwear-ai/uwearaiapp:pr:1486` | Done |
| 6859 | 6830 | `github:uwear-ai/uwear-website:pr:321` | In Review |
| 6860 | 6799 | `github:uwear-ai/uwear-website:pr:309` | In Review |
| 6856 | 6808 | `github:uwear-ai/uwear-backend:issue:2153` | Todo |

The same burst also made 6854 and 6855 (`uwear-website` issues #336 and #335). Those two have no older copy.

## How it was measured

- `illo_read domain.inspect {"domain_id":1,"include_records":true,"limit":100,"format":"compact"}` at 12:08Z. The window runs from `updated_at` 2026-09-17 01:44Z to 2026-09-21 12:02Z.
- Group the records by `(object_key, data.repo, data.number)`. Five keys have two records. In all five pairs, one record has `updated_at` 2026-09-21 12:02:00–12:02:38.
- Control: in the same 100-record window, no duplicate key exists where neither record was written at 12:02Z.

## Why it matters

- Any per-ticket lookup (R1 occupancy veto, R3, the triage handler) can read the wrong copy. A status fix on one copy leaves the other copy stale.
- #292 already fixed "upsert PR records idempotently". This looks like a new write path that does not use that upsert, or a regression.

## Unknown

- R1 cannot see which writer ran at 12:02Z. #900's last comment (2026-09-18) says no Domain-1 projection exists yet. If someone created that projection this morning, it is the likely writer.

## Acceptance

1. Name the writer that made records 6854–6860.
2. That writer upserts on `external_id` (or `repo`+`number`+`object_key`) and makes no copy on a second run.
3. Archive the five copies, or merge them into the older records, with no status change on the older records.

redawear:
Draft PR #906 carries the code part (acceptance 1 and 2).

[...]

PR #906, opened 2026-09-21
fix(domains): stop duplicate tracker records — upsert on external_id and repo+number (#905)

redawear:

Part of #905. #905 stays open after this merges, until the five live copies are archived (acceptance 3). R3 verifies that and then shuts the issue.

## What was wrong

On 2026-09-21 at 12:02Z, five Domain-1 tracker records got a copy with the same `external_id`. Example: record 6857 copies 6793 (`github:uwear-ai/uwearaiapp:pr:1485`).

**The writer (acceptance 1):** Illo Cycle 8, "GitHub Reflex — event fast lane", run 26331 (started 12:00:05Z). The domain event on record 6857 (event 22438) names `run_id` 26331. The run writes through `manage_domain`, which calls `AsyncDomainService.create_record`.

**Why the guard did not stop it:** the upsert guard from #292 keys only on the fields `repo` + `pr_number` (fallback `pr_url`). The live tracker schema has `number`, `url` and `external_id`. So the guard found no key and inserted a new row. The `ticket` object type had no guard at all.

## What this PR does (acceptance 2)

`create_record` now finds the same entity by:
1. `external_id`, when the object type declares that field (`github:` ids compare without case), else
2. `repo` + `number` (or the legacy `pr_number`), with a `url` / `pr_url` fallback for `/pull/` and `/issues/` links.

One pure function, `_select_same_entity_records`, owns these rules. A `repo` + `number` match never joins two records that carry different external ids.

The merge rules from #292 stay the same: the lowest id is the kept record, empty fields fill from the copies, the incoming data wins for the fields it sends, and the extra copies are archived. Object types with neither field keep plain create.

## The five live copies (acceptance 3)

This PR does not change live data. After the deploy, the next `create_record` call for each key merges the pair into the older record and archives the copy. A writer that updates a record by id does not trigger the merge, so if a copy is still active a day after the deploy, archive it by hand. Records 6854 and 6855 have no older copy and stay as they are.

## How to test after the deploy

1. Wait for the next GitHub Reflex run (every 15 minutes), or submit a tracker write for `github:uwear-ai/uwearaiapp:pr:1485`.
2. Run `illo_read domain.inspect {"domain_id":1,"record_id":6857,"include_events":true}`. Expect `archived_at` set and a reason that says "Merged duplicate record into record 6793".
3. Read record 6793. Expect its `status` unchanged (`Done`).
4. Group the newest 100 Domain-1 records by `external_id`. Expect no key with two active records.

## Evidence

- Fast suite (the CI command, no path argument): **5334 passed, 11 skipped, 0 failed** on `38faa0c1`. The first commit `8efcbe60` gave 5331 passed, 0 failed.
- `tests/test_domains_service.py`: 89 passed. New cases: tracker PR and ticket by `external_id` and by `repo` + `number`, existing copies merge, conflicting ids stay apart, non-tracker object types, and a plain object type still creates.
- The Postgres lane (`requires_db`) did not run: no local Postgres. The fix pass removed the only SQL change, so the Postgres path now uses the same query as before this PR.
- Code-quality review (Codex, thermo-nuclear rubric): the first pass was BLOCKING on split identity rules. Commit `38faa0c1` fixes that. Re-review on `38faa0c1`: **APPROVE**, the finding is resolved, no new findings.


redawear:

Superseded by the consolidated **[#910](https://github.com/Illospace/illospace/pull/910)**, which carries this branch plus the other two open R3 PRs. One merge click instead of three, and the merged tree has been gated as a tree: **5358 passed, 0 failed, 0 errors** at 21:27–21:29Z on machine load 5.4.

All three branches are pairwise disjoint in files and none touches a migration, so this is not a conflict fix; it is one decision instead of three.

Leaving this PR open until #910 merges, so nothing is lost if that merge is reverted.

